PR #55, opened 2020-09-22
Fix the Any problem in codegen classes.  Adds additional post-processing to the codegen script

fsuits:
## Types of changes

- \[x \] Bug fix (non-breaking change which fixes an issue)
- \[ x\] Breaking change (fix or feature that would cause existing functionality to change)
- \[ x\] My code follows the code style of this project.

This script relies on datamodel-code-generator version 0.5.28, which produces class that will work but are not strict.  The fix_any.py script replaces the Any references appropriately and reorders the classes to avoid forward references.  

#closes 12


scripts/fix_any.py
```diff
@@ -0,0 +1,130 @@
...
+# class to the singular form.
+# It then reorders the classes so there are no forwards required.
+# This reordering is optimistic and assumes there is nothing circular involved -
+# and that there are no side effect classes that become forwarded.
+
+import os
+import sys
+
+pattern1 = 'ies: Optional[Dict[str, Any]]'
+pattern2 = 's: Optional[Dict[str, Any]]'
```
butler54, line 16: Should we have a catch pattern which is just Dict[str, Any] to test for when we can potentially escape it.

butler54 (approved):
LGTM

brunomarq (approved):
LGTM

PR #60, opened 2020-09-23
feat: implement trestle init command and update specs

### LINKED ISSUE #36: Implement `trestle init`

Implement `trestle init` according to [specifications](https://github.com/IBM/compliance-trestle/blob/develop/docs/trestle-spec.md).

## Completion Criteria

Complete implementation and testing.

## Expected Behavior

- Directory structure created for trestle project according to specification.

brunomarq:
## Types of changes

closes #36


tests/data/split_merge/.trestle/config.ini
```diff
@@ -5,29 +5,29 @@ targets = trestle.oscal.target.Target
 target-control-implementations = trestle.oscal.target.TargetControlImplementation
 
 [catalog]
-decomposition = ['groups[].controls[]']
+decomposition_rules = ['catalog.groups.*.controls.*']
 number_of_groups = 2
```
butler54, line 9: I'm not sure I understand the enumerations here by the names. I think you need to change the variable names to be explicitly 'init' based. E.g. haow would these variables be used by different commands? (or confuse the users because of it).

trestle/resources/config.ini
```diff
@@ -0,0 +1,33 @@
+[key-model-mapping]
+groups = trestle.oscal.catalog.Group
```
butler54, line 2: I don't understand what this is and why we would be codifying the mapping here in a config file. This should be in source as it's not user configurable.

i'm also confused as to why you would need anything beyond the top level models.

trestle/core/commands/merge.py
```diff
@@ -0,0 +1,24 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Trestle Merge Command."""
+
+from ilcli import Command
+
+
+class MergeCmd(Command):
```
butler54, line 21: A quick question do we want to have all of these commands in their own module or not? it seems very 'java'y and unnecessary. Happy to discus.

tests/data/split_merge/step3-split_metadata_array_additionalproperties_elements/catalogs/mycatalog/next_step.sh
```diff
@@ -0,0 +1,6 @@
+# Splits the properties listed in the -e option that must exist in the file specified in the -f option.
+trestle split -f ./groups.json -e 'groups.*.controls.*'
```
butler54, line 2: We may need to discuss globbing behaviour here at some point.

tests/trestle/test_cli.py
```diff
@@ -0,0 +1,33 @@
...
+
+import pytest
+
+from trestle import cli
+
+
+def test_run():
+    """Test cli call."""
+    testargs = ['trestle']
+    with patch.object(sys, 'argv', testargs):
```
butler54, line 29: This is nice. I wonder whether we should do this as an integration test phase? as we might need to setup some quite elaborate directories in the future. E.g. do we register a pytest annotation so we can differentiate between UT's and tests which will impact the file system.

trestle/cli.py
```diff
@@ -12,14 +12,41 @@
...
+from trestle.core.commands.merge import MergeCmd
+from trestle.core.commands.replicate import ReplicateCmd
+from trestle.core.commands.split import SplitCmd
+from trestle.core.commands.validate import ValidateCmd
+
+
+class Trestle(Command):
+    """Manage OSCAL files in a human friendly manner."""
+
+    subcommands = [InitCmd, CreateCmd, SplitCmd, MergeCmd, ReplicateCmd, AddCmd, ValidateCmd, ImportCmd, AssembleCmd]
```
butler54, line 34: See the comment I had above w.r.t a single module. I think we could also do a recursive import rather than having to maintain a list (but that's just been tricky and does not need to be solved now).

trestle/core/commands/create.py
```diff
@@ -0,0 +1,86 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Trestle Create Command."""
+
+from ilcli import Command
+
+
+class CatalogCmd(Command):
```
butler54, line 21: As above. I think we are doing even for this command a lot of 'special' behaviour and hard coding of OSCAL models. If we create a dict (maybe in the trestle.oscal module) which is something like

```python
top_level_models = {
'catalog': catalog.Catalog,
'profile': profile.Profile
...
}
I think we may be able to compress this down signficantly. The objective really is to have less boiler plate than is required. 

This won't be generically applicable, however, I would think would be very useful.

File: trestle/core/commands/init.py [hunk 8]
```diff
@@ -0,0 +1,66 @@
...
+        """Create the directory tree if it does not exist."""
+        # Prepare directory list to be created
+        directory_list = [const.TRESTLE_CONFIG_DIR]
+        for model_dir in const.TRESTLE_MODEL_DIRS:
+            directory_list.append(model_dir)
+            directory_list.append(os.path.join('dist', model_dir))
+
+        # Create directories
+        for directory in directory_list:
+            os.makedirs(name=directory, mode=0o7777, exist_ok=True)
```
butler54, line 60: Should we be specifiying the mode or inheriting it based on user config?

trestle/core/const.py
```diff
@@ -15,9 +15,20 @@
 # limitations under the License.
 """Core constants module containing all constants."""
 
-DIR_TRESTLE = '_trestle'
-DIR_TRESTLE_BACKUP = f'{DIR_TRESTLE}/backup'
-DIR_TRESTLE_DIST = f'{DIR_TRESTLE}/dist'
+TRESTLE_MODEL_DIRS = [
```
butler54, line 18: In reference to the constant dict I previously mentioned - I think we can exploit these lists much better.

trestle/core/parser.py
```diff
@@ -122,12 +121,11 @@ def parse_file(file_name: str, model_name: str):
     if file_name is None:
         raise TrestleError('file_name is required')
 
-    with open(file_name) as f:
-        data = yaml.load(f, yaml.FullLoader)
-        rkey = root_key(data)
-        if model_name is None:
-            model_name = to_full_model_name(rkey)
-        return parse_dict(data[rkey], model_name)
+    data = fs.load_file(file_name)
```
butler54, line 124: Created issue #62 to have a look at this file.

I also wondering whether there are any benefits in using the inbuilt parsers from pydantic.

trestle/resources/config.ini
```diff
@@ -0,0 +1,33 @@
...
+controls = trestle.oscal.catalog.Control
+targets = trestle.oscal.target.Target
```
butler54, line 4: Somewhere in this file we should have a preference for output (json vs yaml). I would like to default to json for now as it's less lossy.

trestle/resources/config.ini
butler54: This would be as everything else should be inferrable based on which of the top level models you are acting on.

butler54 (changes requested):
I carpet bombed comments through - there are a few specific improvements I annotated, however, my biggest concern is that we have two areas of strange behaviour:
1. Lots of duplication in referencing models names as strings which is really messy. I know we will have it but we should try to minimize it.
2. Sometimes we infer lists of things. Sometimes we do not. We should try to be consistent.

tests/data/split_merge/.trestle/config.ini
brunomarq: The default decomposition rules would be used in `trestle create` and `trestle import` commands only.
We could potentially use them as well in `trestle split` if no arguments are passed to a `trestle split` command and if it is run from the base of a model project. But I would avoid that because of the number of 
if's which can cause more confusion to the user.

trestle/resources/config.ini
brunomarq: Remove key-model-mapping from config.ini. It is indeed inferrable.

trestle/core/commands/merge.py
brunomarq: The purpose of having each command in their own module is to increase code cohesion. It feels to me that having all commands in the same module would inevitably lead to low cohesion.

trestle/core/commands/init.py
brunomarq: Removed mode= parameter and relying of current umask.

trestle/core/parser.py
butler54: I'm going to make this as resolved for now and to be deal with with issues relating to trestle.core.parser

trestle/core/commands/create.py
brunomarq: Created issue #64.

trestle/resources/config.ini
brunomarq: Created issue #64.

trestle/core/const.py
brunomarq: Created issue #64.

trestle/resources/config.ini
```diff
@@ -0,0 +1,27 @@
+[catalog]
+decomposition_rules = ['catalog.groups.*.controls.*']
+number_of_groups = 2
```
butler54, line 3: Seem to have lost the comment here - do we want to change 'number_of_groups' to 'create_number_of_groups' or similar?

butler54 (approved):
LGTM - see the single comment.

trestle/resources/config.ini
brunomarq: done

PR #105, opened 2020-10-06
Extend functionality for detecting duplicate values in pydantic objects

fsuits:
## Types of changes

- \[ x] New feature (non-breaking change which adds functionality)
- \[ x] My code follows the code style of this project.
- \[ x] I have added tests to cover my changes.
- \[ x] All new and existing tests passed.

Add functionality and tests for duplicate values of specified type in pydantic objects.  Now have two ways to check: via variable name, e.g. uuid, or by variable type, e.g. target.Prop


fsuits:
Please hold off approving because I want to hook this into the cli.  I will update shortly.

tests/trestle/core/utils_test.py
```diff
@@ -44,3 +47,41 @@ def test_get_elements():
...
+
+    yaml_path = pathlib.Path('tests/data/yaml')
+
+    # test with valid pydantic target
+    good_target_path = yaml_path / 'good_target.yaml'
+    good_target = ostarget.TargetDefinition.oscal_read(good_target_path)
+    assert mutils.has_no_duplicate_values_generic(good_target, 'uuid')
+
+    # test with pydantic target containing duplicates
+    bad_target_path = yaml_path / 'bad_target.yaml'
```
butler54, line 66: I wonder whether we should make our 'bad_object' files describe why they are bad?

trestle/core/utils.py
```diff
@@ -44,6 +44,88 @@ def get_elements_of_model_type(object_of_interest, type_of_interest):
...
+        else:
+            next_item = next(o_iter, None)
+            while next_item is not None:
+                new_list = find_values_by_name_generic(next_item, var_name)
+                if new_list:
+                    loe.extend(new_list)
+                next_item = next(o_iter, None)
+            return loe
+    return loe
+
```
butler54, line 79: So this one I'm a little concerned that this is going to do some slightly crazy things depending on type. I understand that there is an issue with the current 'finding' of objects of interest function which I wrote. This I'm worried that we are implicitly iterating over pydantic internals and I'm not sure what could happen.

would a 1/2 way approach be to twist the original search function to work by field name?

butler54 (changes requested):
See the attached comments I'm a bit worried finding values by name might generate some craziness particularly if the name collides with a python internal (such as class).

Happy to discuss further.

trestle/core/utils.py
fsuits: I have one method that traverses object based on pydantic type - and it is similar to yours but in both of my methods I explicitly use the object's iterator to go through it.

But for the specific task of finding duplicate uuid's, the uuid's are simply string members of the class and I need to find them by name - i.e. use the generic search rather than pydantic.  I guess an alternative would be to scan a file like target.py and find all types that contain a uuid.  There are some actual uuid types like LocationUuid, but other uuid's are simply members of a class, like in Prop.  I would need to build a list of such objects - and search for all things like LocationUuid and Prop.  I now realize I am not catching LocationUuid's so I need to do that  - but accessing Prop.uuid would require building the list of classes that contain uuid.

butler54: Lets discuss in the standup.

butler54:
You might want to let me know when you are good again.

fsuits:
This PR is now ready for review and merge if it is ok.

trestle/core/commands/validate.py
```diff
@@ -51,21 +49,29 @@ def _init_arguments(self):
...
+        if args.item is None:
+            raise TrestleError(f'Argument "-{const.ARG_ITEM_SHORT}" is required')
+        item = args.item
 
-        try:
-            split_plan.execute()
-        except Exception as ex:
-            split_plan.rollback()
-            raise TrestleError(f'Could not perform operation: {ex}')
+        model = ostarget.TargetDefinition.oscal_read(pathlib.Path(args.file))
```
leninmehedy, line 65: Please load the file content using the utility function. Example code can be found below:
https://github.com/IBM/compliance-trestle/blob/develop/trestle/core/commands/split.py#L66

leninmehedy (approved):
Looks good.

butler54 (approved):
LGTM

PR #133, opened 2020-10-12
Merge latest code into develop

leninmehedy:
We need to merge our latest code into develop to keep develop branch upto date.
Let's DO NOT `Squash & Merge` this since we would like to preserve the history of individual commits from multiple people for better tracking of the changes.


brunomarq (approved):
LGTM

PR #169, opened 2020-10-19
docs: change of trestle model directory structure

### LINKED ISSUE #162: Update trestle specs: change to directory structure of a trestle model project

Currently, `trestle create/split/merge/add/remove` commands rely on a directory structure where a directory representing a model object such as `metadata` contains a special file that represents the main content of that model type such as `metadata.json`. That directory can also contain other files or subdirectories depending on whether its main model had been split further. Take the following tree structure as an example:

```
.
├── nist800-53
│   ├── catalog.json
│   ├── groups
│   │   ├── 00000__group
│   │   │   ├── controls.json
│   │   │   └── group.json
│   │   ├── 00001__group
│   │   │   ├── controls.json
│   │   │   └── group.json
│   │   └── groups.json
│   └── metadata
│       ├── metadata.json
│       ├── responsible-parties.json
│       └── roles.json
```

This current structure might be a bit confusing to both user and developer in the sense that every folder always have a special file that represents the main model plus, potentially, other files/subdirectories. During a merge operation, these other files/subdirectories cannot simply be merged with the main model because the root element in the json files are not all at the same level as the main model. For example, `metadata.json` has a root element called `metadata` and `responsible-parties.json` has a root element called `responsible-parties`. When merging `responsible-parties` with `metadata`, it would be incorrect for the resulting file to contain 2 root keys of `metadata` and `responsible-parties`. The correct behaviour would be to have `responsible-parties` placed under `metadata` element.

It would also be more intuitive for a user to look at a subdirectory and expect everything under that subdirectory to be subelements of the model represented by that subdirectory name.

So, we propose a new directory structure where the json file representing the main model is always one level up in relation to a split folder. For example, `metadata.json` would be at the same level as `metadata` directory as opposed to under it.
An example of this new directory structure corresponding to the example above is:

```
.
├── nist800-53.new
│   ├── catalog.json
│   └── catalog
│       ├── metadata.json
│       ├── metadata
│       │   ├── roles.json
│       │   └── responsible-parties.json
│       └── groups
│           ├── 00000__group.json
│           ├── 00000__group
│           │   └── controls.json
│           ├── 00001__group.json
│           └── 00001__group
│               └── controls.json
```

Notice that in the new structure, expanded arrays such as `groups` do not have a corresponding `groups.json` file. In the old structure that file would be there with its contents being an empty json array that should never be edited by the user. To avoid the chance of a user editing a file that is not supposed to be edited, we decided to remove those types of files. The same applies to expanded named fields (aka dicts).

## Completion Criteria

- https://github.com/IBM/compliance-trestle/blob/develop/docs/trestle-spec.md updated with all examples in the new directory structure
- update examples under `test` directory

brunomarq:
## Types of changes

- \[ \] My code follows the code style of this project.

- fixes #162


docs/trestle-spec.md
```diff
@@ -1,4 +1,4 @@
-# Trestle Specifications (v0.0.1)
+# Trestle Specifications (v0.1.0)
```
butler54, line 1: Can we make this the CLI spec explicitly

tests/data/split_merge/step0-merged_catalog/next_step.sh
```diff
@@ -1,9 +1,11 @@
+cd $TRESTLE_BASEDIR/catalogs/mycatalog
+
 # Splits the properties listed in the -e option that must exist in the file specified in the -f option.
-trestle split -f ./catalog.json -e 'catalog.uuid,catalog.metadata,catalog.groups,catalog.back-matter'
+trestle split -f ./catalog.json -e 'catalog.metadata,catalog.groups,catalog.back-matter'
 
 # In the near future, trestle split should be smart enough to figure out which json file contains the elemenets you want to split
 # In that case, the -f option would be deprecated and the commands would look like:
-trestle split -e 'uuid,metadata,groups,back-matter'
+trestle split -e 'catalog.metadata,catalo.groups,catalog.back-matter'
```
butler54, line 8: Typo here `catalo.groups`

butler54 (approved):

PR #173, opened 2020-10-19
Stop generation of timestamp in oscal files, other cleanup

fsuits:
## Types of changes

- \[x \] My code follows the code style of this project.
- \[x \] All new and existing tests passed.

Remove timestamps from generated oscal files to avoid diffs on regen.
Regenerating will no longer cause diffs.
The option to disable timestamps is not in our version of datamodel-codegen so it happens in fix_any.py.
conlist is imported only as needed when a special Union appears.
Add r to regex strings to mark them as raw strings.
Flake8 still complains apparently due to false positives FS003 and F722


butler54 (approved):
LGTM - I noticed the file header has been stripped back too.

PR #216, opened 2020-11-10
Validation of duplicates now uses object factory

fsuits:
## Types of changes

- \[ \] All commits are signed.

Created generic object factory for registration of arbitrary validators.
First implemented validator is a check for duplicate items in an oscal object.
Several file names changed to drive convention for additional validators added later.
Some Typing added but more is needed.


trestle/core/duplicates_validator.py
```diff
@@ -0,0 +1,60 @@
...
+
+        file_path = pathlib.Path(oscal_object_name).absolute()
+        model_type, _ = fs.get_contextual_model_type(file_path)
+        model: OscalBaseModel = model_type.oscal_read(file_path)
+
+        loe = find_values_by_name(model, item_name)
+        if loe:
+            nitems = len(loe)
+            return 0 if nitems == len(set(loe)) else 1
+        return 0
```
butler54, line 46: See above also on documentation.

trestle/core/commands/validate.py
```diff
@@ -48,33 +45,10 @@ def _init_arguments(self) -> None:
...
-        if args.file is None:
-            raise TrestleError(f'Argument "-{const.ARG_FILE_SHORT}" is required')
+        args_raw = args.__dict__
 
-        if args.mode is None:
-            raise TrestleError(f'Argument "-{const.ARG_MODE_SHORT}" is required')
-        mode = args.mode
-        if mode != const.VAL_MODE_DUPLICATES:
-            raise TrestleError(f'Mode value "{mode}" is not recognized.')
+        validator = validator_factory.create(args_raw[const.ARG_MODE])
```
butler54, line 52: So one issue I have here is I have no way from the documentation you provide to discover what are valid modes?

The other question I have here is in the scenario you have provided there is an arbitrary list of arguments that can be provided and a user has no way to discover what they are. I'm presuming we will want to have some global arguments that are constant such as file and 

Finally I would think we would want to allow multiple modes to be passed in (via a comma separated list?).

trestle/core/duplicates_validator.py
```diff
@@ -0,0 +1,60 @@
...
+
+import pathlib
+from typing import Any, Dict
+
+from trestle.core.base_model import OscalBaseModel
+from trestle.core.validator_helper import find_values_by_name
+from trestle.utils import fs
+
+
+class DuplicatesValidator:
```
butler54, line 26: Would it make sense for this to inherit from an abstract base class the method interfaces for validate?

This would avoid the need for a builder per class as we would use the abstract base class to define the interface. It would also allow you to have a minimal set of typing as well the trade off would be that object factory is a completely generic factory which you use registation to build a specific instance.

trestle/core/duplicates_validator.py
butler54, line 46: Failures are going to be problematic here as you have no try: catch: to deal with exceptions anywhere I can see.

trestle/core/duplicates_validator.py
```diff
@@ -0,0 +1,60 @@
...
+    """Find duplicate items in oscal object."""
+
+    def __init__(self, **kwargs: Dict[str, Any]) -> None:
+        """No initialization needed."""
+        pass
+
+    def validate(self, **kwargs: Dict[str, Any]) -> int:
+        """Perform the validation."""
+        oscal_object_name = kwargs['file']
+        item_name = kwargs['item']
```
butler54, line 36: Not sure how you feel about just passing the args object then you can just access items via 

args.file and args.item

butler54 (changes requested):
There appear to be gaps in the testing (at least I cannot see whether we are testing expectations for the factory).

The biggest question I have right now is we are totally unconstrained in terms of the pattern which we expect validators to follow and I have no way as a user of discovering this information.

trestle/core/duplicates_validator.py
fsuits: Right now that is just an implementation example and I'm not sure the current best way to load a module.  The main thing is that it receives kwargs and outputs 0 or 1 - and it is a specific validation mode: duplicates.

trestle/core/commands/validate.py
fsuits: Again this is early and I only have one example validation mode.  It is now easier to add others.  I'm not even sure "file name" makes sense here.
The main thing is that I am assuming there will be different validation "modes" - and each will have a registered factory object.  And all the info needed to perform it will come from kwargs.

trestle/core/duplicates_validator.py
fsuits: If we can define an interface for validators that would be fine.  Right now I am leaving it open ended since I don't know the range of scenarios or use cases.

trestle/core/duplicates_validator.py
fsuits: Again that is just a specific implementation and I assume it will change with how we load models including distributed.

trestle/core/duplicates_validator.py
fsuits: That would be fine - I just took Dict as more generic.

butler54 (approved):
LGTM

PR #219, opened 2020-11-11
Fix/add

aNebula:
## Types of changes

- \[x] All commits are signed.

`trestle add` now works with partial models, which have already been split. This also prevents adding same element to partial models which previously been split from the parent model.


butler54 (approved):
LGTM
Are there addtional tests we should add to double check the bug does not exist?

PR #226, opened 2020-11-13
Trestle create command

butler54:
## Types of changes

- \[x\] All commits are signed.
 
Completed first pass implementation of trestle create. Additional functionality (like default splits) could be added, however, have not been yet.

### DIFF
--- a/trestle/core/generators.py
+++ b/trestle/core/generators.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 """Capabilities to allow the generation of various oscal objects."""
 import uuid
-from datetime import datetime
+from datetime import date, datetime
 from enum import Enum
 from typing import Any, Optional, Type, Union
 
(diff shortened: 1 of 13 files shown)


aNebula (commented):
LGTM

aNebula (approved):
LGTM

trestle/core/generators.py
```diff
@@ -0,0 +1,102 @@
...
+    """
+    # FIXME: Should be in separate generator module as it inherits EVERYTHING
+    if type_ is datetime:
+        return datetime.now().astimezone()
+    elif type_ is bool:
+        return False
+    elif type_ is int:
+        return 0
+    elif type_ is str:
+        return 'REPLACE_ME'
```
fsuits, line 51: There is a python faker module that looks ok - and it makes uuids.

fsuits (approved):
Looks good - I just had some comments.  I will use it as an example so let me know if anything is not intended as a forward looking approach

PR #269, opened 2020-12-17
Feature/trestle tasks transform osco

### LINKED ISSUE #260: Transform OSCO yaml into NIST OSCAL-like Assessment Results Observations list json as a trestle task

Assumptions:

- input directory contains a collection of OSCO produced yaml files
- input directory contains, optionally, a metadata file comprising additional reference data named _oscal-metadata.yaml_
- output directory exists or is creatable

Approach:

- create the output directory, if necessary
- for each input OSCO yaml file:
    - discover the rule and result pair entries contained within
    - construct an observation for each entry, including any additional related information from the metadata file if present
    - create the corresponding output file, only if the file does not already exist or overwrite=true (the default)

Example CLI:
`trestle task osco-to-oscal -c /home/degenaro/git/degenaro.evidence-locker/osco/demo-osco-to-oscal.config`

Example config:
```
[task.osco-to-oscal]

input-dir =  /home/degenaro/git/degenaro.evidence-locker/osco/input
output-dir = /home/degenaro/git/degenaro.evidence-locker/osco/output
output-overwrite = true
```

degenaro:
## Types of changes

- \[x]  New feature (non-breaking change which adds functionality)
- \[x] All commits are signed.

## Description

- Provide transformation function from dict of OSCO YAML (XCCDF) to dict of OSCAL-like JSON (observations list).
- Accept transformation function metadata YAML to enhance produced OSCAL-like JSON.
- Provide trestle task _osco-to-oscal_ to read input OSCO files from input directory and write OSCAL-like transformation data to corresponding files in output directory.
- Provide test-cases covering transformation function and task.

## Comments

The OpenShift Compliance Operator (see https://docs.openshift.com/container-platform/4.6/security/compliance_operator/compliance-operator-understanding.html) employs OpenSCAP, a NIST-certified tool, to scan and enforce security policies. The scan results are available as YAML. The osco-to-oscal function and task perform a transformation from this YAML format to NIST OSCAL-like JSON as a list of observations, each observation conforming to the _observation_ component of Assessment Results (see https://pages.nist.gov/OSCAL/documentation/schema/assessment-results-layer/assessment-results/json-schema/#oscal-ar-json_observation).

 `closes #260`

### DIFF
--- a/tests/data/tasks/osco/input/oscal-metadata.yaml
+++ b/tests/data/tasks/osco/input/oscal-metadata.yaml
@@ -1,5 +1,7 @@
  
 ssg-ocp4-ds-cis-10.221.139.104-pod:
+   locker: https://github.ibm.com/degenaro/evidence-locker
+   namespace: xccdf
    subject-references:
       component:
          uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
@@ -16,6 +18,8 @@ ssg-ocp4-ds-cis-10.221.139.104-pod:
             cluster-region: us-south
             
 ssg-ocp4-ds-cis-10.221.139.105-pod:
+   locker: https://github.ibm.com/degenaro/evidence-locker
+   namespace: xccdf
    subject-references:
       component:
          uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
@@ -32,6 +36,8 @@ ssg-ocp4-ds-cis-10.221.139.105-pod:
             cluster-region: us-south
             
 ssg-ocp4-ds-cis-10.221.139.107-pod:
+   locker: https://github.ibm.com/degenaro/evidence-locker
+   namespace: xccdf
    subject-references:
       component:
          uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
@@ -48,6 +54,8 @@ ssg-ocp4-ds-cis-10.221.139.107-pod:
             cluster-region: us-south
          
 ssg-rhel7-ds-cis-10.221.139.104-pod:
+   locker: https://github.ibm.com/degenaro/evidence-locker
+   namespace: xccdf
    subject-references:
       component:
          uuid-ref: 89cfe7a7-ce6b-4699-aa7b-2f5739c72001
@@ -64,6 +72,8 @@ ssg-rhel7-ds-cis-10.221.139.104-pod:
             cluster-region: us-south
             
 ssg-rhel7-ds-cis-10.221.139.105-pod:
+   locker: https://github.ibm.com/degenaro/evidence-locker
+   namespace: xccdf
    subject-references:
       component:
          uuid-ref: 89cfe7a7-ce6b-4699-aa7b-2f5739c72001
@@ -80,6 +90,8 @@ ssg-rhel7-ds-cis-10.221.139.105-pod:
             cluster-region: us-south
             
 ssg-rhel7-ds-cis-10.221.139.107-pod:
+   locker: https://github.ibm.com/degenaro/evidence-locker
+   namespace: xccdf
    subject-references:
       component:
          uuid-ref: 89cfe7a7-ce6b-4699-aa7b-2f5739c72001
--- a/tests/data/tasks/osco/input/oscal-metadata.yaml
+++ b/tests/data/tasks/osco/input/oscal-metadata.yaml
@@ -1,0 +1,97 @@
+ 
+ssg-ocp4-ds-cis-10.221.139.104-pod:
+   subject-references:
+      component:
+         uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
+         type: component
+         title: Red Hat OpenShift Kubernetes
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d3e
+         type: inventory-item
+         title: Pod
+         properties:
+            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-0000026b.iks.ibm
+            cluster-name: ROKS-OpenSCAP-1
+            cluster-type: openshift
+            cluster-region: us-south
+            
+ssg-ocp4-ds-cis-10.221.139.105-pod:
+   subject-references:
+      component:
+         uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
+         type: component
+         title: Red Hat OpenShift Kubernetes
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d4f
+         type: inventory-item
+         title: Pod
+         properties:
+            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-000001fe.iks.ibm
+            cluster-name: ROKS-OpenSCAP-1
+            cluster-type: openshift
+            cluster-region: us-south
+            
+ssg-ocp4-ds-cis-10.221.139.107-pod:
+   subject-references:
+      component:
+         uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
+         type: component
+         title: Red Hat OpenShift Kubernetes
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d4e
+         type: inventory-item
+         title: Pod
+         properties:
+            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-00000338.iks.ibm
+            cluster-name: ROKS-OpenSCAP-1
+            cluster-type: openshift
+            cluster-region: us-south
+         
+ssg-rhel7-ds-cis-10.221.139.104-pod:
+   subject-references:
+      component:
+         uuid-ref: 89cfe7a7-ce6b-4699-aa7b-2f5739c72001
+         type: component
+         title: RedHat Enterprise Linux 7.8
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d3e
+         type: inventory-item
+         title: VM
+         properties:
+            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-0000026b.iks.ibm
+            cluster-name: ROKS-OpenSCAP-1
+            cluster-type: openshift
+            cluster-region: us-south
+            
+ssg-rhel7-ds-cis-10.221.139.105-pod:
+   subject-references:
+      component:
+         uuid-ref: 89cfe7a7-ce6b-4699-aa7b-2f5739c72001
+         type: component
+         title: RedHat Enterprise Linux 7.8
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d4f
+         type: inventory-item
+         title: VM
+         properties:
+            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-000001fe.iks.ibm
+            cluster-name: ROKS-OpenSCAP-1
+            cluster-type: openshift
+            cluster-region: us-south
+            
+ssg-rhel7-ds-cis-10.221.139.107-pod:
+   subject-references:
+      component:
+         uuid-ref: 89cfe7a7-ce6b-4699-aa7b-2f5739c72001
+         type: component
+         title: RedHat Enterprise Linux 7.8
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d4e
+         type: inventory-item
+         title: VM
+         properties:
+            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-00000338.iks.ibm
+            cluster-name: ROKS-OpenSCAP-1
+            cluster-type: openshift
+            cluster-region: us-south
+   
--- a/tests/data/tasks/osco/input/oscal-metadata.yaml
+++ b/tests/data/tasks/osco/input/oscal-metadata.yaml
@@ -16,43 +16,7 @@ ssg-ocp4-ds-cis-10.221.139.104-pod:
             cluster-name: ROKS-OpenSCAP-1
             cluster-type: openshift
             cluster-region: us-south
-            
-ssg-ocp4-ds-cis-10.221.139.105-pod:
-   locker: https://github.ibm.com/degenaro/evidence-locker
-   namespace: xccdf
-   subject-references:
-      component:
-         uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
-         type: component
-         title: Red Hat OpenShift Kubernetes
-      inventory-item:
-         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d4f
-         type: inventory-item
-         title: Pod
-         properties:
-            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-000001fe.iks.ibm
-            cluster-name: ROKS-OpenSCAP-1
-            cluster-type: openshift
-            cluster-region: us-south
-            
-ssg-ocp4-ds-cis-10.221.139.107-pod:
-   locker: https://github.ibm.com/degenaro/evidence-locker
-   namespace: xccdf
-   subject-references:
-      component:
-         uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
-         type: component
-         title: Red Hat OpenShift Kubernetes
-      inventory-item:
-         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d4e
-         type: inventory-item
-         title: Pod
-         properties:
-            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-00000338.iks.ibm
-            cluster-name: ROKS-OpenSCAP-1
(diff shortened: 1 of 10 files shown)


degenaro:
Ready for review.

trestle/lib/osco.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def _get_target(self, root):
+        """Extract 'target' info from the stringified XML."""
+        target = None
+        for lev1 in root:
+            tag = _remove_namespace(lev1.tag)
+            if tag == 'target':
+                target = root.find(lev1.tag).text
+                break
+        return target
```
butler54, line 254: ElementTree has search functionality which may be easier  / faster to use than iteration.

I'm slightly confused as to the _remove_namespace function as I could not see where that existed in your test dataset. If that is a requirement I think we need a test explicitly for this (and obviously test data).

trestle/lib/osco.py
```diff
@@ -0,0 +1,297 @@
...
+    def _get_config_map(self, root):
+        """Extract 'config_map' info from the stringified XML."""
+        retval = ''
+        for lev1 in root:
+            tag = _remove_namespace(lev1.tag)
+            if tag == 'benchmark':
+                href = lev1.get('href')
+                retval = href.split('.')[0].rsplit('/', 1)[1]
+                break
+        return retval
```
butler54, line 265: Is this a config map or something else? I think this is the benchmark ID we are removing here.

trestle/lib/osco.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def _get_result(self, lev1):
+        """Extract 'result' info from the stringified XML."""
+        result = None
+        for lev2 in lev1:
+            tag = _remove_namespace(lev2.tag)
+            if tag == 'result':
+                result = lev1.find(lev2.tag).text
+                break
+        return result
```
butler54, line 275: Similar comment to the above w.r.t. using Elementree internal search so we don't need to iterate outselves.

trestle/lib/osco.py
```diff
@@ -0,0 +1,297 @@
...
+
+    def _post(self, config_map, idref, time, result_type, target, name):
```
butler54, line 277: OCD kicking in - Function name and description are a little off.

trestle/lib/osco.py
```diff
@@ -0,0 +1,297 @@
...
+        properties = [p1, p2, p3, p4]
+        evidence['properties'] = properties
+        return evidence
+
+    def _create_property(self, ns, classification, name, value):
+        prop = {}
+        if ns is not None:
+            prop['ns'] = ns
+        prop['class'] = classification
+        prop['name'] = name
```
butler54, line 122: Please use the trestle.oscal.assessment_results classes here.

trestle/lib/osco.py
```diff
@@ -0,0 +1,297 @@
...
+                component = entry['subject-references']['component']
+                inventory_item = entry['subject-references']['inventory-item']
+                subject_references = [component, inventory_item]
+        except Exception:
+            logger.debug('missing or invalid subject references')
+        return subject_references
+
+    def _create_observation_methods(self, rule, metadata):
+        observation_methods = ['TEST-AUTOMATED']
+        return observation_methods
```
butler54, line 141: this function has input variables which are being ignored.

trestle/tasks/osco_to_oscal.py
```diff
@@ -0,0 +1,184 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class OscoToOscal(TaskBase):
+    """
+    Task to convert OSCO yaml to OSCAL json.
+
+    Attributes:
+        name: Name of the task.
```
butler54, line 40: So a question: Is OSCO the name for the compliance operator or one that we invented internally?

I think we probably need to spell it out in the comments.

trestle/tasks/osco_to_oscal.py
```diff
@@ -0,0 +1,184 @@
...
+                idir = pathlib.Path(self._config.get('input-dir'))
+                if idir is None:
+                    logger.error(f'config missing "input-dir"')
+                    return TaskOutcome('failure')
+                odir = pathlib.Path(self._config.get('output-dir'))
+                if odir is None:
+                    logger.error(f'config missing "output-dir"')
+                    return TaskOutcome('failure')
+                overwrite = self._config.getboolean('output-overwrite', True)
+                quiet = self._config.getboolean('quiet', False)
```
butler54, line 118: Trestle task should observe the verbosity flag. What that implies is if a user passes -V to the command line it will set the log level to print out debug. I think it would make more sense to align with this paradigm.

trestle/tasks/osco_to_oscal.py
```diff
@@ -0,0 +1,184 @@
...
+                os.makedirs(odir, exist_ok = True)
```
butler54, line 119: you can use the pathlib internals which would be odir.mkdir(exist_ok=True, parents=True). While it is no cleaner we are trying to focus on using the pathlib abstractions.

trestle/tasks/osco_to_oscal.py
```diff
@@ -0,0 +1,184 @@
...
+                    return TaskOutcome('simulated-failure')
+                overwrite = self._config.getboolean('output-overwrite', True)
+                for ifile in sorted(pathlib.Path(idir).iterdir()):
+                    parts = ifile.parts
+                    ifn = parts[len(parts)-1]
+                    if ifn.endswith('oscal-metadata.yaml'):
+                        continue
+                    if ifn.endswith('oscal-metadata.yml'):
+                        continue
+                    ofile = self._calculate_ofile(ifn, odir)
```
butler54, line 83: The help information is not explicit about these particular files are required. If so we should state it in the help info print.

trestle/tasks/osco_to_oscal.py
```diff
@@ -0,0 +1,184 @@
...
+                    if not overwrite:
+                        if os.path.exists(ofile):
```
butler54, line 129: you can use ofile.exists() here instead.

trestle/tasks/osco_to_oscal.py
```diff
@@ -0,0 +1,184 @@
...
+                            logger.error(f'file exists: {ofile}')
+                            return TaskOutcome('failure')
+                    mfile = idir / 'oscal-metadata.yaml'
```
butler54, line 132: This line is inconsistent with the assumptions in line 123

trestle/tasks/osco_to_oscal.py
```diff
@@ -0,0 +1,184 @@
...
+                    if not quiet:
+                        logger.info(f'Rules Analysis:')
+                        logger.info(f'config_maps: {analysis["config_maps"]}')
+                        logger.info(f'dispatched rules: {analysis["dispatched_rules"]}')
+                        logger.info(f'result types: {analysis["result_types"]}')
+                return TaskOutcome('success')
+            logger.error(f'config missing')
+            return TaskOutcome('failure')
+        except Exception:
+            traceback.print_exc()
```
butler54, line 148: use traceback.format_exc() combined with the logging level of your choice.

butler54 (changes requested):
So I've requested a bunch of changes - some are bigger some are smaller stylistic changes due to internal choices within the project.

Bigger ones are:
1) Adoption of the trestle models internally.
2) I've got some confusion in the xml parsing mostly due to the fact it's the first time i've looked at the code

I did not get to extensive testing yet.

Issues not covered in file by file comments:
1) For UT's make sure we are doing individual UT's for functions. Some of the functions are covered by testing higher up the stack, however, exception handling layers in between can mean this leaves you open to strange behaviour.
2) It would be good if code typing annotations to function definitions and even better (for public method) argument / return docstrings.

i'll note on that last point we have not yet been consistent with the trestle codebase and this is something that is evolving. A first pass over the code base with annotations would probably be enough.

degenaro:
> Bigger ones are:
> 
>     1. Adoption of the trestle models internally.
> 
>     2. I've got some confusion in the xml parsing mostly due to the fact it's the first time i've looked at the code

With respect to the first big issue, the osco.py module was purposefully made stand-alone, depending only on python standard modules, with the indent of making it portable without requiring other parts of trestle. If this is a false design point, then perhaps the location in "lib" is questionable too.

trestle/tasks/osco_to_oscal.py
degenaro: Fixed.

trestle/tasks/osco_to_oscal.py
degenaro: Fixed.

trestle/tasks/osco_to_oscal.py
degenaro: Fixed.

tests/data/tasks/osco/input/oscal-metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+   subject-references:
+      component:
+         uuid-ref: 56666738-0f9a-4e38-9aac-c0fad00a5821
+         type: component
+         title: Red Hat OpenShift Kubernetes
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d3e
+         type: inventory-item
+         title: Pod
+         properties:
```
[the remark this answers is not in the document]
degenaro: Fixed. Used github.mycorp.com as exemplar.

tests/data/tasks/osco/input/oscal-metadata.yaml
```diff
@@ -0,0 +1,37 @@
...
+      component:
+         uuid-ref: 89cfe7a7-ce6b-4699-aa7b-2f5739c72001
+         type: component
+         title: RedHat Enterprise Linux 7.8
+      inventory-item:
+         uuid-ref: 46aADFAC-A1fd-4Cf0-a6aA-d1AfAb3e0d3e
+         type: inventory-item
+         title: VM
+         properties:
+            target: kube-br7qsa3d0vceu2so1a90-roksopensca-default-0000026b.iks.ibm
```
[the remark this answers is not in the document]
degenaro: Fixed. Used "mycorp" as exemplar.

tests/data/tasks/osco/input/ssg-ocp4-ds-cis-10.221.139.104-pod.yaml
```diff
@@ -0,0 +1,689 @@
+apiVersion: v1
+data:
+  exit-code: "2"
+  results: |
+    <?xml version="1.0" encoding="UTF-8"?>
+    <TestResult xmlns="http://checklists.nist.gov/xccdf/1.2" id="xccdf_org.open-scap_testresult_xccdf_org.ssgproject.content_profile_cis" start-time="2020-08-03T02:26:26+00:00" end-time="2020-08-03T02:26:26+00:00" version="0.1.52" test-system="cpe:/a:redhat:openscap:1.3.3">
+              <benchmark href="/content/ssg-ocp4-ds.xml" id="xccdf_org.ssgproject.content_benchmark_OCP-4"/>
+              <title>OSCAP Scan Result</title>
+              <profile idref="xccdf_org.ssgproject.content_profile_cis"/>
+              <target>kube-br7qsa3d0vceu2so1a90-roksopensca-default-0000026b.iks.ibm</target>
```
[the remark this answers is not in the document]
degenaro: Fixed. Used "mycorp" and "111.222.333.444" for corp and IP respectively.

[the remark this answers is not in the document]
degenaro: Fixed. Used "mycorp" and "111.222.333.444" for corp and IP respectively.

tests/data/tasks/osco/output/ssg-ocp4-ds-cis-10.221.139.104-pod.oscal
```diff
@@ -0,0 +1,7254 @@
+{
```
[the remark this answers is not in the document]
degenaro: Fixed. Suffix is .json now.

trestle/tasks/osco_to_oscal.py
degenaro: Not sure what you mean here by inconsistent? This is checking if the output file already exists and if so overwriting only if permitted. Previous lines (e.g., 123) are concerned with input, I believe, and and therefore not germane.

butler54: Okay reading it again - It looks to me like 122 and mfile point to the same location (unless I am wrong) which would imply that the metadata will never get read if it exists. Unless I am missing something.

butler54:
@degenaro I think the Q here is over how python packages are assembled. Typically there is no concept of lib within a project - just external libs via pypi (excluding c bindings and other things).

With that I think we have a few choice we can go down:
1. We build it as an independent mini library for that particular piece (I can help set it up for you). Under that circumstance this PR would add it as a dependency to trestle and the wrapping function on the trestle side would do the validation of oscal.
2. We refactor to meet trestle expectations.

So some thoughts on above:
1. Refactoring into trestle does not prevent the library from being called directly - it does mean they would do so via calling `trestle.lib.osco` (or similar). - probably something closer to `trestle.tasks.osco` but can be debated.
2. The downside of course is code bloat (pulling all of trestle to get the osco conversion functionality)
3. Independent libraries of course will need to be maintained and Q/A'ed - i'm not sure how this would work w.r.t. the auditree discussion.

degenaro:
re: there is no concept of lib within a project

If you recall, the evolution of this was: 1) extend OSCO to produce OSCAL directly, which was abandoned; then 2) create a "standalone library" that does the transformation. It seems to be that we should do the integration with trestle as you suggest, meaning employ the oscal modules already in place; then sell trestle as the "standalone library". If something without trestle is still wanted, we can create a separate project as you suggest. I think we need to find out who the customers of such a project would be, no?

tests/data/tasks/osco/output/ssg-ocp4-ds-cis-10.221.139.104-pod.oscal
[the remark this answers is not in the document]
degenaro: Fixed.

trestle/tasks/osco_to_oscal.py
degenaro: Not sure what is inconsistent? There are 3 kinds of files: 1) input osco yaml 2) input oscal metadata 3) output oscal. 

1) are the input files to be converted from OSCO formatted yaml into OSCAL formatted json
2) is optional an yaml file that, if present, augments the produced OSCAL formatted json
3) are output json files, one for each of the input yaml files in 1)

If should be noted that if the output file to be produced already exists, then it will not be replaced unless the config specifies overwrite = True, which is the default. That is specify overwrite = False in the config to prevent overwriting.

tests/data/tasks/osco/input/oscal-metadata.yaml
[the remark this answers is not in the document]
degenaro: Not sure what you mean by "double check these repos?". I've updated the oscal-metadata.yaml file to use mycorp.com. I've left in "degenaro" for userid in the example url. Happy to change if that is undesirable. 

The data contained within this file are **metadata** and optional. These data are used to provide improved partial results. For example, the locker does not need to actually exist. Nothing is fetched from it. This locker string (e.g. "https/...) is simply copied into the partial results.

tests/data/tasks/osco/input/oscal-metadata.yaml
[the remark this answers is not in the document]
degenaro: Changed to "mycorp".

tests/data/tasks/osco/input/ssg-ocp4-ds-cis-10.221.139.104-pod.yaml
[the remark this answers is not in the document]
degenaro: My claim is that de-identification changes have now been delivered and is complete.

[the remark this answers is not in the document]
degenaro: My claim is that de-identification changes have now been delivered and is complete.

tests/data/tasks/osco/output/ssg-ocp4-ds-cis-10.221.139.104-pod.oscal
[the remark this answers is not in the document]
degenaro: Agreed. Output file is now: osco-pod-oscal.json.

trestle/tasks/osco_to_oscal.py
degenaro: My claim is that updated code and comments now address your "inconsistent with the assumptions" concern.

trestle/tasks/osco_to_oscal.py
degenaro: Fixed. Added required/optional designation in the text.

degenaro:
> Issues not covered in file by file comments:
> 
>     For UT's make sure we are doing individual UT's for functions. Some of the functions are covered by testing higher up the stack, however, exception handling layers in between can mean this leaves you open to strange behaviour.
>     It would be good if code typing annotations to function definitions and even better (for public method) argument / return docstrings.

With respect to UT's, there is only one public function and there is a unit test for that. My claim is that each of the public methods has a unit test. Presently, osco_to_oscal.py has 100% coverage and osco.py has 94% coverage.

With respect to typing, recently added are annotations for public methods and function.

degenaro:
Closing this PR. Will open a new one based on the OSCAL 1.0 develop branch.

PR #279, opened 2021-01-05
Fixes issues with the generators for constrained ints.

butler54:
The fix should be relatively robust. 

I've included a bunch of other fixes that should stabilize your codebase a bit as well.

Be aware that trestle.oscal.* files will most likely conflict with yours due to my hack-changes which you can freely overwrite.


fsuits (approved):
Looks good will merge into my branch

PR #280, opened 2021-01-05
Allow generators to handle cases of List[NonOscalObject] as a field.

butler54:


fsuits (approved):
Looks good will merge

PR #283, opened 2021-01-06
More UT coverage

butler54:
Covers:
- cmd_utils_test
- fs_test
- partial load_distributed_test


fsuits (approved):
Looks good will merge

PR #300, opened 2021-01-15
Make fix_any deterministic in output generation

fsuits:
fix_any.py previously generated files that may have slightly different class output order.  It should now be deterministic.   In this first iteration with the new version there are big changes to the files, but after this they should be constant unless there is a change either to the nist source or the behavior of datamodel-codegenerator.

Also added full coverage for snake and camel tests

## Types of changes


butler54 (approved):
LGTM

PR #311, opened 2021-01-22
Enhancement to handle arboretum fetcher-built OSCO evidence as input

### LINKED ISSUE #310: Transform OSCO json produced by Arboretum fetcher into NIST OSCAL-like Assessment Results Observations list json file(s) as a trestle task

Assumptions:

    input directory contains one or more OSCO Arboretum fetcher produced cluster resource json files
    input directory contains, optionally, a metadata file comprising additional reference data named oscal-metadata.yaml
    output directory exists or is creatable

Approach:

    create the output directory, if necessary
    for each input OSCO cluster resource json file:
        for each ConfigMap:
            discover the rule and result pair entries contained within
            construct an observation for each entry, including any additional related information from the metadata file if present
            create the corresponding output file, only if the file does not already exist or overwrite=true (the default)

Example CLI:
trestle task osco-to-oscal -c /home/degenaro/git/degenaro.evidence-locker/osco/demo-osco-to-oscal-arboretum.config

Example config:

[task.osco-to-oscal]

input-dir =  /home/degenaro/git/degenaro.evidence-locker/osco-arboretum/input
output-dir = /home/degenaro/git/degenaro.evidence-locker/osco-arboretum/output
output-overwrite = true

## References

See Arboretum https://github.com/ComplianceAsCode/auditree-arboretum/issues/9 cluster resource fetcher.

degenaro:
## Types of changes

# Description

- Provide trestle task osco-to-oscal to read input Arboretum-Auditree fetcher-built files from input directory and write OSCAL-like transformed data to corresponding files in output directory.
- Provide test-cases covering transformation function and task.

# Comments

The Compliance-As-Code auditree-arboretum cluster resource fetcher (see https://github.com/ComplianceAsCode/auditree-arboretum/issues/9) produces evidence which comprises the equivalent of one or more OpenShift Compliance Operator (OSCO) .yaml files.

In the end, the XML results of the OSCO .yaml file and of the auditree-arboretum fetcher are formatted the same. Thus, ingesting either OSCO .yaml files or fetcher .json files boils down to being able to find the XML results then transform them into OSCAL. The code contributed here provides the ability to do the OSCAL transformation on the fetcher-built .json files.

closes #310


trestle/tasks/osco_to_oscal.py
```diff
@@ -119,33 +120,30 @@ def simulate(self) -> TaskOutcome:
...
+                    # create the OSCAL .json file from the OSCO and the optional osco-metadata files
+                    observations, analysis = osco.get_observations(collection[oname], metadata)
+                    # write the OSCAL to the output file
+                    self._write_content(ofile, observations, True)
+                    # display analysis
+                    if not quiet:
+                        logger.debug(f'[simluate] Rules Analysis:')
+                        logger.debug(f'[simluate] config_maps: {analysis["config_maps"]}')
+                        logger.debug(f'[simluate] dispatched rules: {analysis["dispatched_rules"]}')
+                        logger.debug(f'[simluate] result types: {analysis["result_types"]}')
```
butler54, line 146: So one item I forgot to mention. The logger should automatically observe a -v flag passed into the CLI which will turn on debug logging. If that behaves as expected that would be my preference (as it removes the need for an additional flag.

It is not a blocker more of a FYI

butler54 (approved):
LGTM
If you can have a quick look at the comment w.r.t. logging before merging it would be appreciated.

trestle/tasks/osco_to_oscal.py
degenaro: For the most part the code does expect and follow the standard of the -v flag. Note that the task produces logging output about the file written and some analysis of the results encountered at the info level of execute. The quiet flag is meant to suppress all logging output, meaning the aforementioned info. At this point it is a frill, you might say, and can be removed entirely...or not. Nevertheless and with respect to symmetry, for the case of simulate these outputs are at the debug level while at the execute level they are at the info level. I think it's OK the way it is, but am not wedded to the need for the quiet flag. If preference is to remove, just say the word. Is there a way to suppress .info level output already?

PR #315, opened 2021-01-26
Feat/assemble

aNebula:
`trestle assemble` implemented with UT

## Types of changes

- \[x\] All commits are signed-off.


tests/trestle/core/commands/assemble_test.py
```diff
@@ -0,0 +1,67 @@
...
+# limitations under the License.
+"""Tests for trestle assemble command."""
+import pathlib
+import shutil
+import sys
+from unittest import mock
+
+from trestle.cli import Trestle
+from trestle.oscal.catalog import Catalog
+from trestle.utils.load_distributed import load_distributed
```
butler54, line 24: So food for thought - I wonder whether with some of these modules we should expose interior functions up to the parent module level. 

e.g. `trestle.utils.load_distributed` lets discuss at some point (no need to change here.)

trestle/core/commands/assemble.py
```diff
@@ -15,10 +15,189 @@
...
+from trestle.core.models.file_content_type import FileContentType
+from trestle.core.models.plans import Plan
+from trestle.oscal import assessment_plan
+from trestle.oscal import assessment_results
+from trestle.oscal import catalog
+from trestle.oscal import component
+from trestle.oscal import poam
+from trestle.oscal import profile
+from trestle.oscal import ssp
+from trestle.oscal import target
```
butler54, line 37: Giveen there is a common pattern of importing everything we should find a simpler way to do this (it should be viable in one line).

butler54 (changes requested):
Mostly good please:
1) Review comments for 'nice to haves'
2) Complete UT coverage. It looks like we are not covering exception criteria.
3) For public docstrings can you please start documenting args in the style here: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

tests/trestle/core/commands/assemble_test.py
aNebula: I think it'll feel very satisfying to import modules that way!

trestle/core/commands/assemble.py
aNebula: agreed. We now have 3 commands that do this.

butler54 (approved):
LGTM

PR #367, opened 2021-02-19
Added bulk operations for assemble

fsuits:
## Types of changes

Assemble can now perform for all models of specified type.  Currently the Plan created is for each model assembled - rather than for the entire operation.  It can be done for the entire operation, but having it succeed as much as possible is perhaps better.


butler54 (approved):
LGTM

PR #373, opened 2021-02-23
Fix to changes caused by new datamodel-gen, using original nist source

fsuits:
## Types of changes

Changes made to fix_any to fix an existing problem in component.py.  In the process I ran make develop, which installed the latest datamodel-codegen, and ran make gen-oscal, which downloaded the latest nist source (Feb 21, 2021).  Both new items caused problems and it seems the nist-source is very different as of around Feb 12.  So I turned off the updating of the nist module in gen-oscal, and I made new fixes to fix_any.py to accommodate the new dmcg and to fix the bug in component.py.  That bug appears to have fixed itself with the updated version of dmcg, but new issues were introduced.


butler54 (approved):
LGTM - i've had some problems with the github UI.

Issue #384, opened 2021-03-03
AssessmentResults results.local-definitions missing items


degenaro:
**Describe the bug**
Seemingly, "inventory-items" and other keys are missing.

**To Reproduce**
Compare current AssessmentResults class LocalDefintions to [https://pages.nist.gov/OSCAL/documentation/schema/assessment-results-layer/assessment-results/json-model-map/](https://pages.nist.gov/OSCAL/documentation/schema/assessment-results-layer/assessment-results/json-model-map/). See screenshot below.

**Expected behavior**
Was looking to use inventory-items. Have hacked it into my copy of AssessmentResults for now.

**Screenshots**
![Screenshot from 2021-03-03 03-41-57](https://user-images.githubusercontent.com/520029/109783195-b92f4500-7bd7-11eb-8184-aff644fe715c.png)

fsuits:
I confirm this is being dropped by fix_any.  I will look into it.

degenaro:
Fixed by #386.

PR #386, opened 2021-03-04
Fix/fix any drops item and pydantic 1.8.1 update - issues #382 and #384

fsuits:
## Types of changes

Issue #384 fixed with update and cleanup of fix_any to handle *Item classes properly.
Issue #382 fixed by upgrading to pydantic version 1.8.1 and above


trestle/oscal/assessment_plan.py
```diff
@@ -1714,38 +1753,9 @@ class Risk(OscalBaseModel):
...
-    relevant_evidence: Optional[List[RelevantEvidence]] = Field(
-        None, alias='relevant-evidence', min_items=1
-    )
-    remarks: Optional[Remarks] = None
-
-
 class LocalDefinitions(OscalBaseModel):
     components: Optional[Dict[str, SystemComponent]] = None
-    inventory_items: Optional[List[Inventory]] = Field(
+    inventory_items: Optional[List[InventoryItem]] = Field(
```
degenaro, line 1758: I think this is what will fix my issue, and in that case LGTM!

trestle/oscal/assessment_results.py
```diff
@@ -1707,6 +1717,43 @@ class Action(OscalBaseModel):
...
+    )
+    remarks: Optional[Remarks] = None
+
+
+class LocalDefinitions(OscalBaseModel):
+    add_objectives_and_methods: Optional[List[LocalObjective]] = Field(
+        None, alias='add-objectives-and-methods', min_items=1
+    )
+    activities: Optional[List[Activity]] = Field(None, min_items=1)
+    remarks: Optional[Remarks] = None
```
degenaro, line 1754: Should there not be inventory_items here?

scripts/fix_any.py
```diff
@@ -301,17 +251,12 @@ def reorder(class_list):
...
 
 
-def fix_header(header, needs_conlist):
+def fix_header(header):
     """Fix imports and remove timestamp from header."""
     new_header = []
     for r in header:
-        # Any is needed for biblio so still need to import it
-        # r = re.sub(' Any, ', ' ', r) # noqa: E800
         if r.find(' timestamp: ') >= 0:
```
koxudaxi, line 258: @fsuits
There is `--disable-timestamp` option which is disabled generating this line.

trestle/oscal/assessment_results.py
fsuits: As discussed, the oscal schema has local-definitions defined locally twice in separate locations, and they have different contents.  The fix is to allow the creation of two separate classes, LocalDefinitions and LocalDefinitions1 - as created by datamodel-codegen.  The correct one should be used with the corresponding parent class - and pydantic should enforce that agreement.  The fundamental problem appears to be in the oscal schema itself.  This happens in many places so users must be aware of the possibility for multiple classes differing only by the index.

scripts/fix_any.py
fsuits: thank you @koxudaxi.  I tried adding this to the command in gen_oscal.py and the timestamp still appeared.  I do see it in the -h help listing for datamodel-codegen so it appears to be present in the version.  I may not be invoking it properly.

koxudaxi: The option works on my envriomment.
```bash
$ datamodel-codegen  --version                                                       
0.9.0
$ echo "{}"| datamodel-codegen --input-file-type jsonschema --disable-timestamp 
# generated by datamodel-codegen:
#   filename:  <stdin>

from __future__ import annotations
...

```
But, let's ignore my point out. I will check it when I know how to generete `compliance-trestle` models in locally.  Thank you.

trestle/oscal/assessment_results.py
```diff
@@ -1791,41 +1999,15 @@ class Risk(OscalBaseModel):
...
-    actions: Optional[List[Action]] = Field(None, min_items=1)
-    related_controls: Optional[ReviewedControls] = Field(None, alias='related-controls')
-    responsible_roles: Optional[Dict[str, ResponsibleRole]] = Field(
-        None, alias='responsible-roles'
+class LocalDefinitions1(OscalBaseModel):
+    components: Optional[Dict[str, SystemComponent]] = None
+    inventory_items: Optional[List[InventoryItem]] = Field(
+        None, alias='inventory-items', min_items=1
     )
-    remarks: Optional[Remarks] = None
```
degenaro, line 1820: Good that inventory-items appears in LocalDefinitions1. That should fix my current issue. However, according to the website https://pages.nist.gov/OSCAL/documentation/schema/assessment-results-layer/assessment-results/json-model-map/ there should also be users, assessment-assets, and assessment-actions. Perhaps this is because trestle is backlevel version of OSCAL? No worries for me.

degenaro (approved):
Proof is in the pudding. I donwloaded your branch and fit my adjusted changes on top, removed my hack to fix inventory-items, and the trestle task tanium-to-oscal worked as expected.

butler54 (approved):
LGTM

PR #399, opened 2021-03-13
Auto-update pre-commit hooks

aNebula:
Update versions of tools in pre-commit 
configs to latest version


butler54:
Closing - need to remove auto-updating.

PR #454, opened 2021-04-08
feat:Improved serialisation support.

### LINKED ISSUE #453: Provide wrapping functionality such as oscal_write to allow in-memory object wrapping form OscalBaseModel
In order for the workflow for #451 and #452 to be executed successfully we also need a method to serialize in memory rather than just onto disk such as `oscal_write`.

butler54:

Added support for OscalBaseModel to produce serialised output suitable for machine to machine consumption.

Closes #453


fsuits (approved):
LGTM

Issue #452, opened 2021-04-08
Refactor OSCO to OSCAL and tanium to OSCAL to use ResultsTransfomer interface (beneath the task)


butler54:
Based on #451 refactor OSCO to OSCAL and tanium to OSCAL to exploit new `ResultsTransformer` interface.

This implies refactoring the output such that:

1) A single 'blob' is consumed:
    - Where possible this blob can be multiple objects as defined by the transformer  e.g. 
```json
{
       "metadata": "my_metadata",
       "object": "my Object"
}
```

2) The object returned is a `Results` class as defined in #451 

3) If there is no finding associated with the method a single finding will be created (to preserve valid OSCAL) where the UUID maps to a 'all zeros' UUID e.g.:
```json
"finding": {
  "uuid": "00000000-0000-4000-8000-000000000000",
  "title": "no Finding",
  "description": "No finding."
}
```
Title and description are arbitrary.

butler54:
Closed with #466

PR #466, opened 2021-04-13
ResultsTransformer interface

degenaro:
## Types of changes

- \[x] I have updated the documentation accordingly.

## Description

- support blob interface
- only support dir output (not merged file)
- adjust cli doc
- move to folder transforms/implementations/tanium.py
- move to folder transforms/utils/tanium_helper.py
- catch exceptions
- use oscal_write

Addresses 1/2 of issue #452


trestle/transforms/implementations/tanium.py
```diff
@@ -0,0 +1,71 @@
...
+    @staticmethod
+    def set_timestamp(value: str) -> None:
+        """Set the default timestamp value."""
+        datetime.datetime.strptime(value, '%Y-%m-%dT%H:%M:%S%z')
+        TaniumTransformer._timestamp = value
+
+    @staticmethod
+    def get_timestamp() -> str:
+        """Get the default timestamp value."""
+        return TaniumTransformer._timestamp
```
butler54, line 44: @degenaro do you think timestamp is something we should generically make available as part of the base class? I see it as been something that is generic enough it might be worthwhile (as a future enhancement). 

Will not block the PR but feedback would be appreciated.

butler54 (approved):
LGTM 

@fsuits to add transformer registration prior to the PR being merged.

PR #522, opened 2021-05-17
fix: Ensure PR automation system cannot override checks.

butler54:
The current PR automation system allowed merging when the PR had not passed all required tests.


fsuits (approved):
LGTM

PR #546, opened 2021-05-27
fix: move unreachable debug statement

deenine:
## Types of changes

- \[x\] Bug fix (non-breaking change which fixes an issue)
- \[ \] All commits are signed-off.

Debug statement indicating start of split command was before logging level set - therefore the only way this debug statement is displayed is by editing cli.py to set the logging level, or editing utils/log.py to change the default level. Moving debug statement after logging level is set per argument results in this statement being printed when `trestle split -v ....` is specified.


butler54 (approved):
LGTM

PR #570, opened 2021-06-16
feat: Allow explicit inclusion / exclusion of readme files in author workflows

butler54:
## Types of changes

- \[x\] All commits are signed-off.


trestle/core/commands/author/docs.py
```diff
@@ -197,19 +211,49 @@ def _validate_dir(
...
             return 1
         return self._validate_dir(
-            template_file, governed_heading, self.task_path, validate_header, validate_only_header, recurse
+            template_file,
+            governed_heading,
+            self.task_path,
+            validate_header,
+            validate_only_header,
+            recurse,
+            readme_validate
```
fsuits, line 258: Just a comment that these multiple options may be better as a separate class or something.  I had added the validate_header and validate_only_header - which overlap in meaning.  And now with readme_validate there is even more.  Will this possibly have more added?  I think it's fine as it is, but does stand out in the code as a lot of arguments in a call.

fsuits (approved):
LGTM - with one comment that maybe the many arguments regarding the validation options should be encapsulated somehow.

trestle/core/commands/author/docs.py
butler54: I thought about that. I think we should clean up the functionality a bit going forward. validate header only can technically be dropped due to the separate command.

PR #573, opened 2021-06-17
fix: Release to account for errors occurring in trestle author (logging referring to non-existent class variables).

butler54:


fsuits (approved):
LGTM

PR #650, opened 2021-07-28
feat: describe command to describe contents of model files with optional element path

### LINKED ISSUE #409: Implement trestle split --list-available-elements -> trestle describe -f mymodel.json -e metadata.roles
This is described in the split_merge examples but is not implemented.  Provides a list of which elements can be split at a given level.

fsuits:
## Types of changes

New command describe, to describe contents of model files and elements within them.
Updated docs for describe and split.

closes #409


trestle/core/commands/describe.py
```diff
@@ -0,0 +1,149 @@
...
+                logger.warning('The element path for describe must either be omitted or contain at least 2 parts.')
+                return []
+
+            element_paths = utils.parse_element_arg(model, element_path_str)
+
+            sub_model_element = Element(model)
+
+            for element_path in element_paths:
+                sub_model = sub_model_element.get_at(element_path, False)
+                sub_model_element = Element(sub_model)
```
butler54, line 125: I'm a bit confused here. it seems like here you are trying to cater to mulitiplle models, however, lines 127-147 seem to be catering to a singular element path.

butler54 (changes requested):
It seems that there is (unless I am reading it wrong) some inconsistency as to whether one or multiple element paths are allows. 

If it is mean to be singular it would be worth testing that users only pass one element path.

trestle/core/commands/describe.py
fsuits: This is indeed confusing - but that's the way element paths work.  The element_path list is a list of links in a chain - so you are following the links to the final model.  There is only one final entity being described.  This is all leveraging the existing code for parse_element_args - which parses the element path string (only one, possibly long one) into a list of element paths - each path being a pair representing a link in the chain.  No wildcards or commas are allowed.  The good thing is that the element path construct does allow this all to happen without much code.

fsuits: Oh - it is true that I don't think I check for commas in the path - so I will add that.

butler54 (approved):
LGTM

Issue #599, opened 2021-06-29
As a user I would like to be able to generate object with the optional fields included.


butler54:
## Issue description / feature objectives
Currently we have a generator capability, however, that capability is limited for to mandatory fields. The original reason for this is that there a small number of 'self referencing' classes that may result in loops.

As far as we can tell all loops take the form of:
```python
class Singular(OscalBaseModel):
     singulars: Optional[List[Singular] = Field(None, 'the list of singluar objects we want')
```
OR
```python
class Singular(OscalBaseModel):
     singulars: Optional[Dict[str,Singular] = Field(None, 'the dict of singluar objects we want')
```
These usecases should be detectable.

Put a flag into the generators (backwards compatible) which allows a full set of classes to be generated. This will allow more thorough testing of the classes.

Final note: A solution will be required for `Dict[str, Any]` objects and `include-all` in profile (`Optional[Any]`).

PR #690, opened 2021-08-20
feat: Adding capability to allowing generator to generate optional fields.

butler54:

## Types of changes

Allows the generator to generate optional classes. Nice and safe in terms of what it can generate. More complex object can be generated by chaining this functionality together.

The idea would be to introduce this to `trestle add`.

closes: #599

### DIFF
--- a/trestle/core/generators.py
+++ b/trestle/core/generators.py
@@ -80,8 +80,10 @@ def generate_sample_value_by_type(
             return 'REPLACE_ME'
     elif hasattr(type_, '__name__') and 'ConstrainedIntValue' in type_.__name__:
         # create an int value as close to the floor as possible does not test upper bound
-        multiple = type_.multiple_of or 1  # default to every integer
-        floor = type_.ge or type_.gt + 1 or 0  # default to 0
+        multiple = type_.multiple_of if type_.multiple_of else 1  # default to every integer
+        # this command is a bit of a problem
+        floor = type_.ge if type_.ge else 0
+        floor = type_.gt + 1 if type_.gt else floor
         if math.remainder(floor, multiple) == 0:
             return floor
         else:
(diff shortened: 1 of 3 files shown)


trestle/core/generators.py
```diff
@@ -80,8 +80,10 @@ def generate_sample_value_by_type(
             return 'REPLACE_ME'
     elif hasattr(type_, '__name__') and 'ConstrainedIntValue' in type_.__name__:
         # create an int value as close to the floor as possible does not test upper bound
-        multiple = type_.multiple_of or 1  # default to every integer
-        floor = type_.ge or type_.gt + 1 or 0  # default to 0
+        multiple = type_.multiple_of if type_.multiple_of else 1  # default to every integer
```
fsuits, line 83: Looks like conint is only used by protocol via portrange

trestle/core/generators.py
```diff
@@ -94,22 +96,38 @@ def generate_sample_value_by_type(
...
+) -> TG:
+    """Given a model class, generate an object of that class with sample values.
+
+    Can generate optional variables with an enabled flag. Any array objects will have a single entry injected into it.
+
+    Note: Trestle generate will not activate recursive loops irrespective of the depth flag.
+
+    Args:
+        model: The model type provided. Typically for a user as an OscalBaseModel Subclass.
+        optional: Whether or not to generate optional fields.
```
fsuits, line 117: for clarity might be good to call this include_optional instead of optional because it looks a little odd

fsuits (approved):
LGTM - minor comments

trestle/core/generators.py
butler54: +1

butler54: Done.

trestle/core/generators.py
butler54: I guess we need to support this irrespective given the complexity of NIST's changes.

PR #716, opened 2021-09-03
fix: Refactor underlying methods to isolate calls to Path.cwd()

butler54:

## Types of changes

- [] New feature (non-breaking change which adds functionality)
- [ ] All new and existing tests passed.

I need to write some new tests for the `trestle.utils.fs` module but otherwise this should be good to go. 

This PR isolates CWD calls which has the downside that cwd needs to be explicitly passed more, however, removes some bugs that impacted the `trestle.core.repository`

### DIFF
--- a/trestle/utils/fs.py
+++ b/trestle/utils/fs.py
@@ -118,10 +118,9 @@ def get_relative_model_type(relative_path: pathlib.Path) -> Tuple[Type[OscalBase
     model_type, model_alias = utils.get_root_model(module_name)
     full_alias = model_alias
 
-    for i in range(len(model_relative_path.parts)):
-        tmp_path = pathlib.Path(model_relative_path.parts[i])
-        alias = extract_alias(tmp_path)
-        if i > 0 or model_alias != alias:
+    for index, part in enumerate(model_relative_path.parts):
+        alias = extract_alias(part)
+        if index > 0 or model_alias != alias:
             model_alias = alias
             full_alias = f'{full_alias}.{model_alias}'
             if utils.is_collection_field_type(model_type):
@@ -163,7 +162,7 @@ def get_stripped_model_type(
 
     malias = model_alias.split('.')[-1]
     logger.debug(f'not collection field type, malias: {malias}')
-    if absolute_path.is_dir() and malias != extract_alias(absolute_path):
+    if absolute_path.is_dir() and malias != extract_alias(absolute_path.name):
         split_subdir = absolute_path / malias
     else:
         split_subdir = absolute_path.parent / absolute_path.with_suffix('').name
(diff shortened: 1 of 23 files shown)


trestle/utils/fs.py
```diff
@@ -93,30 +93,18 @@ def has_parent_path(sub_path: pathlib.Path, parent_path: pathlib.Path) -> bool:
...
 
-    if root_path is None or project_model_path is None:
-        raise err.TrestleError('Trestle project model not found')
+    Returns:
+        Type of Oscal Model for the provided model
+        Alias of that oscal model.
 
-    relative_path = path.relative_to(str(root_path))
+    """
     project_type = relative_path.parts[0]  # catalogs, profiles, etc
```
fsuits, line 108: I'm not sure how unhappy cases are handled.  What if the first part of path is foo?  Also what if you have catalogs/catalog.json?  i.e. valid object in wrong part of directory.

trestle/utils/load_distributed.py
```diff
@@ -24,40 +24,40 @@
...
-def _load_list(file_path: Path) -> Tuple[Type[OscalBaseModel], str, List[OscalBaseModel]]:
+def _load_list(abs_path: Path, abs_trestle_root: Path) -> Tuple[Type[OscalBaseModel], str, List[OscalBaseModel]]:
     """Given path to a directory of list(array) models, load the distributed models."""
     aliases_not_to_be_stripped = []
     instances_to_be_merged: List[OscalBaseModel] = []
-    file_path = file_path.resolve()
-    collection_model_type, collection_model_alias = fs.get_stripped_contextual_model(file_path)
-
-    for path in sorted(Path.iterdir(file_path)):
+    abs_path = abs_path.resolve()
```
fsuits, line 31: Kinda weird to call resolve on what is labeled an abs_path.  In that sense file_path is better.

trestle/utils/fs.py
```diff
@@ -126,8 +114,7 @@ def get_contextual_model_type(path: pathlib.Path = None) -> Tuple[Type[OscalBase
     full_alias = model_alias
 
     for i in range(len(model_relative_path.parts)):
```
fsuits, line 116: I would rewrite this because of the range(len) and because the inner loop is somewhat opaque

fsuits (approved):
LGTM but some suggestions in comments for improvement.

trestle/utils/fs.py
butler54: Now the project will choke if the incorrect directory is passed. This relative directory always relative to trestle project root directory.

trestle/utils/fs.py
butler54: Cleaned up partially, however, it's inherited and a bit messy still. For not it is good enough for me.

PR #718, opened 2021-09-06
feat: import cache

fsuits:
## Types of changes

Import now uses cache, and the cache system is generalized to handle relative paths.

### DIFF
--- a/tests/trestle/core/remote/cache_test.py
+++ b/tests/trestle/core/remote/cache_test.py
@@ -271,7 +271,7 @@ def test_sftp_fetcher_bad_uri(uri: str, tmp_trestle_dir: pathlib.Path) -> None:
         'https://:{{mypassword}}@github.com/IBM/test/file'
     ]
 )
-def hide_test_fetcher_bad_uri(tmp_trestle_dir: pathlib.Path, uri: str, monkeypatch: pytest.MonkeyPatch) -> None:
+def test_fetcher_bad_uri(tmp_trestle_dir: pathlib.Path, uri: str, monkeypatch: pytest.MonkeyPatch) -> None:
     """Test fetcher factory with bad URI."""
     if 'https' in uri:
         monkeypatch.setenv('myusername', 'user123')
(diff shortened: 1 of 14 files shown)


tests/trestle/core/remote/cache_test.py
```diff
@@ -41,33 +42,42 @@ def as_file_uri(path: str) -> str:
...
-    fetcher = cache.FetcherFactory.get_fetcher(pathlib.Path(tmp_trestle_dir), catalog_file)
-    # Create/update the cache copy
+    catalog_data.oscal_write(catalog_file)
+    saved_data = fs.load_file(pathlib.Path(catalog_file))
+    if relative:
+        catalog_str = f'./catalogs/{cat_name}' if in_trestle else f'../{cat_name}'
+    else:
+        catalog_str = str(catalog_file)
+    fetcher = cache.FetcherFactory.get_fetcher(tmp_trestle_dir, catalog_str)
+    return fetcher, catalog_data, saved_data
```
butler54, line 65: I'm not sure why we are saving then loading data here. The extra read seems unnecessary (even if we need the file). Users can ask for the dict data by  OscalModel.dict(by_alias=True, exclude_unset=True)

trestle/core/remote/cache.py
```diff
@@ -110,11 +110,16 @@ def get_raw(self, force_update=False) -> Dict[str, Any]:
...
         try:
-            return fs.load_file(self._cached_object_path)
-        except Exception as e:
-            logger.error(f'Cannot fs.load_file {self._cached_object_path}')
-            logger.debug(e)
-            raise TrestleError(f'Cache get failure for {self._uri}') from e
+            raw_data = fs.load_file(self._cached_object_path)
+        except Exception:
+            try:
+                # files from NIST seem to need cp1252 encoding
```
butler54, line 116: NIST files are annotated with utf-8 as the extension type

butler54 (approved):
LGTM see the small changes.

tests/trestle/core/remote/cache_test.py
fsuits: This was needed in order to test the raw mode of fetching - but I have removed the extra read and the corresponding test of raw read.

trestle/core/remote/cache.py
fsuits: Oh - that comment should have been removed.  At first I thought NIST had an issue, but the issue was in cache.py where it wrote the file with pathlib.write_text().  That line now has encoding specified and before it did not - causing the problem.

Comment now removed.

Issue #720, opened 2021-09-09
trestle validate fails on split files probably due to load_distribute


fsuits:
## Describe the bug

When controls are split off from a catalog, the catalog fails validation with:
`WARNING:root:1 validation error for Catalog
controls
  extra fields not permitted (type=value_error.extra)`

## To Reproduce

1. Create a catalog with controls
1. Split off the controls
1. Validate the now split catalog
1. Validate does a load_distributed so this should be fine but it give the above error

## Expected behavior

Validation always uses load_distributed so it should make no difference if the file is split or not.

## Environment

- OS: Windows but I doubt that's a factor

butler54:
Can you please provide the set of commands (including any cd's performed) when running this command?

PR #725, opened 2021-09-15
fix:  Correct split merge pathing inconsistencies.

butler54:
## Types of changes

This fixes issues with the validator across all platforms. 
Closes: #720


tests/trestle/core/commands/validate_test.py
```diff
@@ -210,3 +213,30 @@ def test_validate_dup_uuids(sample_component_definition: ComponentDefinition) ->
     sample_component_definition.components[1].control_implementations[0].uuid = sample_component_definition.components[
         0].uuid
     assert not validator.model_is_valid(sample_component_definition)
+
+
+def test_validate_distributed(
+    testdata_dir: pathlib.Path, tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPatch
+) -> None:
+    """Hello."""
```
fsuits, line 221: Perhaps an actual comment here.

trestle/utils/fs.py
```diff
@@ -447,3 +447,31 @@ def allowed_task_name(name: str) -> bool:
...
+    If relative expansions exist (e.g. ../) the final result must still be within the cwd.
+
+    If an absolute path is provided it tests whether the path is within the cwd or not.
+
+    """
+    # Expand user first if applicable.
+    candidate = candidate.expanduser()
+
+    if not cwd.is_absolute():
+        raise TrestleError('Must provide absolute cwd.')
```
fsuits, line 467: This error msg might need more descriptive info if user sees it.

trestle/utils/load_distributed.py
```diff
@@ -56,8 +55,8 @@ def _load_dict(abs_path: Path, abs_trestle_root: Path) -> Tuple[Type[OscalBaseMo
 
 
 def load_distributed(
-    file_path: Path,
-    trestle_root: Path,
+    abs_path: Path,
```
fsuits, line 58: I'm not clear what the codecov msg is referring to, but if _load_dict is calling load_distrib and it isn't being tested that's a concern.

fsuits (approved):
LGTM after consideration of a few comments made.

trestle/utils/fs.py
butler54: Done.

trestle/utils/load_distributed.py
butler54: This is not unexpected. Actually it should have been there for a while now. The changes in OSCAL reduce the use of DICT structures. I'm not sure we are ready to remove this yet, however, if we can it may simplify code in a number of modules.

PR #736, opened 2021-09-24
feat: osco results remove scc_goal_description

degenaro:
## Types of changes

- [ ] I have added tests to cover my changes.
- [ ] All commits are signed-off.

## Description

- remove `scc_goal_description` class (but keep `benchmark_id` property)
- use result `id` rather than `benchmark.id` for `scc_predefined_profile` class
- fix corresponding expected test outputs
- reduce size of rhel7 expected test output


trestle/transforms/utils/osco_helper.py
```diff
@@ -63,6 +63,7 @@ class RuleUse():
 
     def __init__(
```
butler54, line 64: For these classes we can use Pydantic base_models that will automate some of this setup.

butler54: We can leave this for now though. Just a thought to decrease total LoC

butler54 (approved):
LGTM

PR #755, opened 2021-09-30
fix: Ignore hidden files throughout the project

### LINKED ISSUE #747: Fix merge command `trestle/core/commands/merge.py` inconsistent behaviour
## Describe the bug

When using `Trestle` library, installed via `pip`, the `merge` command fails to merge the broken down catalog file after the folder is opened by the user. Opening the folder with broken down catalog leaves the merge command in permanent state where the merge of the earlier separated file is no longer possible and the following error is thrown:

`trestle.core.commands.merge:72 ERROR: Merge failed: Invalid file content type FileContentType.DIRLIKE`

## To Reproduce

Steps to reproduce the behavior (on MacOS):

1. Go to the [tutorial page](https://ibm.github.io/compliance-trestle/tutorials/trestle_sample_workflow/)
2. Complete steps 1 and 2 in a terminal
3. Split the catalog as described in step 3
4. Go to the catalog folder in the Finder
5. Open created after split folder `catalog`
6. Return to the terminal, now try to merge the file as described in step 3
7. Observe the error: `trestle.core.commands.merge:72 ERROR: Merge failed: Invalid file content type FileContentType.DIRLIKE`

## Expected behavior

Opening the folder by the user does not affect the merge command and the file could be merged back.

## Environment

- OS: macOS Big Sur 11.6
- Python version: 3.9.7
- Trestle version: v0.24.0

enikonovad:

## Types of changes

A presence of the hidden files in the folders was breaking some logic in the project including the behaviour of the `merge` command. This change fixes this issue by ignoring the hidden files. A new function `iterdir_without_hidden_files` was added to `fs` module. It is meant to be used instead of `Path.iterdir()` in the places were hidden files (only) should be ignored.

Some tests were added to cover this bug.

Closes #747


tests/trestle/core/commands/merge_test.py
```diff
@@ -226,6 +230,78 @@ def test_merge_everything_into_catalog(testdata_dir, tmp_trestle_dir):
...
+    # Call merge()
+    generated_plan = MergeCmd.merge(Path.cwd(), ElementPath('catalog.*'), tmp_trestle_dir)
+
+    # Assert the generated plan matches the expected plan'
+    assert generated_plan == expected_plan
+
+    if os.name == 'nt':
+        hidden_file.unlink()
+        hidden_file2.unlink()
+    else:
```
fsuits, line 298: I don't think we need to do cleanup like this in tests.  Is there a special need because they are hidden?

fsuits (approved):
LGTM except I think there is cleanup at the end of tests that I don't think is needed.  I would remove that unless for some reason the hidden files require it.

tests/trestle/core/commands/merge_test.py
butler54: All tests are in tmp directories and should be automatically cleaned up by pytest.

enikonovad: Good catch, removed now

PR #771, opened 2021-10-08
feat: allow author edits and update of profile

fsuits:
## Types of changes

New author command: profile.  profile-generate and profile-assemble
given a profile, generate will create markdown for the profile with special editing sections for any Adds that it does.
A user may edit the content and/or add new sections.
Then profile-assemble will load the markdown and replace the adds in the profile with the new content.

In addition, the objective part of a control is written in the same indented and structured format as the statement.

This includes a re-structuring of ControlIO into separate reader and writer classes.

This is a preliminary PR and the documentation needs updating prior to merge, but it is ready for initial review.

### DIFF
--- a/docs/license.md
+++ b/docs/license.md
@@ -1,7 +1,7 @@
 ```text
                                  Apache License
                            Version 2.0, January 2004
-                        http://www.apache.org/licenses/
+                        https://www.apache.org/licenses/
 
    TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
 
@@ -193,7 +193,7 @@
    you may not use this file except in compliance with the License.
    You may obtain a copy of the License at
 
-       http://www.apache.org/licenses/LICENSE-2.0
+       https://www.apache.org/licenses/LICENSE-2.0
 
    Unless required by applicable law or agreed to in writing, software
    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


enikonovad:
I leave the proposed exception handling for CLI commands here. @fsuits @butler54 please take a look, if you agree with the structure we can adopt it more. Here an example of the proposed structure for the author/profile.py:

The proposition is to handle all unexpected exceptions here and return code 1. We also preserve stack-trace for debug purposes. 
```
def _run(self, args: argparse.Namespace) -> int:
        try:
            trestle_root = args.trestle_root
            if not fs.allowed_task_name(args.output):
                logger.warning(f'{args.output} is not an allowed directory name')
                return 1

            profile_path = trestle_root / f'profiles/{args.name}/profile.json'

            markdown_path = trestle_root / args.output

            return self.generate_markdown(trestle_root, profile_path, markdown_path)
        except Exception as e:
            logger.error(f"Generation of the profile failed with error: {e}")
            logger.debug(traceback.format_exc())
            return 1
```

example of handling error in methods such as `@staticmethod
    def assemble_profile(
        trestle_root: pathlib.Path, orig_profile_name: str, md_name: str, new_profile_name: str
    ) -> int:`

Proposition is that we only catch the exceptions that we expect, we let all other unexpected exceptions fall up the call tree. 
```
...
new_prof_dir = trestle_root / f'profiles/{new_profile_name}'
try:
    if new_prof_dir.exists():
        logger.info('Creating profile from markdown and destination profile directory exists, so deleting.')
        shutil.rmtree(str(new_prof_dir))
    
    new_prof_dir.mkdir()
    orig_profile.oscal_write(new_prof_dir / 'profile.json')

    return 0
except OSError as e:
    raise TrestleError(f'Error while writing profile from markdown to {new_prof_dir}: {e}')
except AnyOtherSpecificError as e:
    raise TrestleError(f"something went wrong: {e}")
```

fsuits:
@enikonovad @butler54 The suggestions for exception handling make sense - I will add them in here and aim to use them in the future.

PR #787, opened 2021-10-13
feat: Support for SSP writing with jinja templating.

### LINKED ISSUE #780: Validate whether Jinja can handle OSCAL style substitution templates
In implementing a (potential) templating mechanism we need to understand how it interacts with the *current* oscal templates. Note that the templating approach taken by OSCAL today is a bit different to what is used in classical templating mechanism like moustache. (see here: https://github.com/usnistgov/OSCAL/issues/971) 

GIven this we must understand the compatibility of this with a presumed templating library (jinja2). The questions include:

1. Given an oscal style expression (e.g.`{{ insert: param, ac-2_prm_1 }}`) can this directly be used for a parameter substitution?
    1. This means that the inner statement `insert: param, ac-2_prm_1` must be mapped into a variable name.
1. If this does not work can we verify that this statement will *not* cause issues for jinja or is isolateable
2. If no issues are found validate that text that is both processed and not processed with Jinja can have a clena parameter substitution.

## Completion Criteria
- Test cases built to test the code are in functionality tests module
- Associated documentation in the issue describing the best approach.

butler54:
## Types of changes

- [ ] I have added tests to cover my changes.

## Summary
- Currently the skeleton of jinja templating code. including ssp_io module.

Closes #780 

## Key links:


fsuits (approved):
LGTM

docs/assets/sample_ssp.md.jinja
```diff
@@ -0,0 +1,37 @@
...
+
+<!-- This iterates over the controls in a resolved catalog and prints out the control statement, summary table,
+and implemented requirements in a format similar to a FedRAMP .docx
+
+Note that the catalog interface has controls properly sorted so is useful to present the controls how they are organised
+in the original catalog 
+-->
+# Control responses and guidance
+{% for group in catalog_interface.get_all_groups_from_catalog() +%}
+## {{ group.title }} {{ group.class }} \({{ group.id|upper }}\)
```
fsuits, line 23: keep in mind some catalogs will have controls not in a group, and the group name for them will be ''

docs/trestle_author_jinja.md
```diff
@@ -0,0 +1,118 @@
...
+```
+
+### SSP interface.
+
+If `-ssp` is passed a variable within the jinja template called `ssp_md_writer` is made available which is an instance of `trestle.core.ssp_io.SSPMarkdownWriter`.
+`-ssp` requires that `-p` has also been set.
+
+This as allows users, as an example to print out a control response, as markdown
+
+```jinja
```
fsuits, line 96: typo: this as

trestle/core/catalog_interface.py
```diff
@@ -226,6 +228,75 @@ def get_count_of_controls_in_catalog(self, recurse: bool) -> int:
...
+        """Get statement label if given."""
+
+        def does_part_exists(part: common.Part) -> bool:
+            does_match = False
+            if part.name and part.name in {'statement', 'item'} and part.id == statement_id:
+                does_match = True
+            return does_match
+
+        control = self.get_control(control_id)
+        label = None
```
fsuits, line 253: get_control returns None if the control_id doesnt exist so I would check that

trestle/core/catalog_interface.py
```diff
@@ -226,6 +228,75 @@ def get_count_of_controls_in_catalog(self, recurse: bool) -> int:
...
+        found_part = None
+        if control.parts:
+            for part in control.parts:
+                # Performance OSCAL assumption, ids are nested so recurse only if prefix
+                if part.id and statement_id.startswith(part.id):
```
fsuits, line 258: I would use as_list on control.parts because they are optional

trestle/core/catalog_interface.py
```diff
@@ -226,6 +228,75 @@ def get_count_of_controls_in_catalog(self, recurse: bool) -> int:
...
+            for new_group in group.groups:
+                for res in CatalogInterface._get_groups_from_group(new_group):
+                    yield res
+
+    @staticmethod
+    def get_label(object_with_props: BaseModel) -> str:
+        """Get the label from an object with properties (such as a control)."""
+        label = ''
+        if object_with_props.props:
+            for prop in object_with_props.props:
```
fsuits, line 294: a control may not have props so unless you're sure the object has props at this point I would check first

trestle/core/ssp_io.py
```diff
@@ -0,0 +1,330 @@
...
+        """
+        if self._ssp is None:
+            raise TrestleError('Cannot get responsible roles, SSP is not set.')
+
+        for impl_requirement in self._ssp.control_implementation.implemented_requirements:
+            if impl_requirement.control_id == control_id:
+                if impl_requirement.responsible_roles:
+                    role_ids = []
+                    for resp_role in impl_requirement.responsible_roles:
+                        role_ids.append(resp_role.role_id)
```
fsuits, line 124: When I create roles from the markdown header and put them in the json ssp, I need to make them ncname compatible, which means spaces get converted to _.  There is no great way to do this, but for the final markdown it might be good just to convert all _ to spaces at this point so they look good.

trestle/core/ssp_io.py
```diff
@@ -0,0 +1,330 @@
...
+        return ''
+
+    def _parameter_table(self, control_id: str, level: int) -> str:
+        """Print Param_id | Default (aka label) | Value or set to 'none'."""
+        if not self._ssp:
+            raise TrestleError('Cannot get parameter table, set SSP first.')
+
+        writer = ControlIOWriter()
+        control = self._catalog_interface.get_control(control_id)
+
```
fsuits, line 155: watch for None here

trestle/core/ssp_io.py
```diff
@@ -0,0 +1,330 @@
...
+        control_impl_req = self._control_implemented_req(control_id)
+        if not control_impl_req:
+            logger.info(f'No implemented requirements found for the control {control_id}')
+            return ''
+
+        md_writer = MDWriter(None)
+        if control_impl_req.statements:
+            for statement in control_impl_req.statements:
+                statement_id = statement.statement_id
+                label = statement_id
```
fsuits, line 223: recommend as_list for statements and parts

fsuits (approved):
LGTM with suggested changes.  All are small but some are important to avoid Exceptions.

PR #797, opened 2021-10-19
feat: Add centralised markdown API

### LINKED ISSUE #744: As a developer I would like a fit for purpose markdown api for trestle.
Mdparsers for python appear to come in two varieties (from what I have seen):
1. Straight to HTML / XML / etc
2. For internal manipulation only (e.g. mdformat)
3. Generating an AST.

So far within trestle two approaches have been taken
1. Using Mistune's AST 
2. Using plain old python to parse the document.
Neither are spectacular in that
1. Mistune's AST is more complicated than trestle's needs.
   2. This is as the output of trestle is typically markdown itself. We are considering the structure of the markdown document, demarked by headings, not whether or not a piece of text has been converted to bold type. 
2. The python code is getting scatteded.
   3. This brings risk of inconsistency.

### Requirements for trestle parsing markdown.
1. Check that markdown does not violate commonmark / is renderable. This is primarily a check that none of the content contains invalid HTML tags within it. It is assumed that HTML will be a rendering engine for any markdown at some point in the future.
2. IDEALLY check whether it fits GitHub flavoured markdown.
3. Be able to identify and build a hierarchy of headings where the 'heading level' (aka the number of `#` symbols) dictates the nesting level.
    1. Expose the headings via an iterator
    2. Provide an interface either iterate through headers in a 'flat' or 'hierarchical manner'
4. Presuming a header which has nested components be able to either:
    1. Retrieve all the content below the header as one blob.
    2. Return the text/ list content directly below the header and the content in the nested sub headings separately.
5.  For a markdown content blob (e.g. a set of lines) be able to identify an unnumbered list
    1. For that list be able to create a nested structure to represent nested lists.
6. For headings be able to identify trestle defined filtering tags to remove content.
7. Be able to run jinja over the content to resolve include tags
8.  Be able to run jinja over the content to be able to insert variables using jinja / moustache

### Anticipated future requirements
1. Be able to identify a table and extract it (presuming only one table exists under a heading)
2. Be able to load the table into a dataframe or list based object.

## Caveats / Assumptions
- The biggest assumption here is that we have two vehicles for delivering content
   - A heading (based) chunk of markdown is associated with a multiline markup field in OSCAL
   - Another markdown document (e.g. an aggregation)

## Current open questions:
- Whether to leverage mistune or similar to create the capability or build from scratch.

## Completion Criteria
- Independent MD parsing class / interface is created in trestle 
- MD parsing is used across all of trestle author (at a minimum).

enikonovad:
## Types of changes

## Summary
At the moment the markdown-related functionality is spread across the project. This change introduces new centralised markdown API that brings the that code together. At the moment only reading and validation is supported, however in the future work we can add markdown writer here as well.

The `MarkdownAPI` object wraps around the markdown processor and validator and provides a singular point of access to all markdown functionality. 

`MarkdownProcessor` constructs a tree-like structure out of the markdown body (excluding the YAML header), and provides an easy access to its contents. Each node in the tree is defined by the key which is a section header and value which is all text inside that section including the inner subsections. `MarkdownNode` functionality allow to access the necessary nodes by the keys. When fetching node by key, all the contents of that section will be returned in form of a subtree.

## Key links:

Closes #744


enikonovad:
`ControlIOReader` still partially uses its own markdown reader to read controls. It might be better to separate it to another ticket once this change gets merged as it requires a lot of work. Please let me know what do you think.

enikonovad:
Added documentation and updated api reference

PR #820, opened 2021-10-29
feat: sample generic oscal transform to filter ssp by profile

### LINKED ISSUE #811: Proposed framework for oscal to oscal transformations
Trestle tasks have been construed for unique tasks which may include a large number of parameters related to content that is external. The complexity of this has relied on a configuration file approach to perform various transformations.

When we consider OSCAL files **exclusively** the complexity is more limited and may be containable to a relatively straight forward command.

The objective here would be (as far as is practical) allow chaining of the transformations. Under this the suggestion is of the form:
trestle transform {input top level model type} --input {potentially multiple instances} --output {single output model} --transform x --transform y --transform {order of operations specified by order provided} --extra-params.

Under this regime all parameters required for the transforms would be globally scoped across all transforms (e.g. if a profile is provided it is consistently used). This will limit the depth of transform that is executable, however should contain the complexity.

### A note on files authored using trestle author in markdown
- This issue presumes that users of this functionality must first convert the markdown to oscal then run the transformation tasks.
- It may be possible to cross link this functionality as well.

## Transforms 
Plase note names are indicative
### filter-by-profile:
- Filters the input document by the controls / parts of a corresponding resolved profile
- Applies too (in decreasing order of priority:
  - SSP
  - Component Def
  - AP
  - AR 
  - POAM
- Is a 1:1 map of the same type for input to output
  - Required a `--profile` to be passed

### filter-exclude-components
- Removes components specified by a --components argument and returns the same object to the output location
- Applies to:
   - Component definition
   - SSP
   - AP
   - AR
   - POAM
### filter-include-components
- Filters exclusively for components specified by a --components argument and returns the same object to the output location
- Applies to:
   - Component definition
   - SSP
   - AP
   - AR
   - POAM

### jijna
Applies to all documents where the document model is processed through a jinja engine. It presumes that the context for templates is the trestle root directory and users can pass:
- Input `include` files
- binary vars that will be true when passed e.g `--jinja-var=foo`
- A limited set of pre-built jinja vars that can be extrapolated and documented (e.g. current time)
- Requires special handling of OSCAL parameter substitutions

### aggregate
- many to one aggregation of a given document type with precedence

### resolve profile
- Runs the profile resolution process
-  input is one profile 
- output is one catalog

#### component-defintion-to-ssp
- Best effort conversion from a component definition to a ssp
#### ssp-to-ap
Best effort conversion of a ssp to an ap

#### ap-ar
Best effort conversion of an ap to an ar.

#### cd-profile
Use the component definition to configure a 'all controls used and configured' profile

#### ssp -> component definition
Convert a ssp to a vendor statement (potentially with property based filter)

## Caveats / Assumptions
- Transformations are exclusively based on the assumption we are handling a top level object.
- 
- One to many transforms are not supported, run mulitple workflows.
- if mutliple inputs are provided an aggregation operation is required

### LINKED ISSUE #854: Allow all trestle author *-assemble commands to do an 'in place update' so that uuids / other data is not changed.

fsuits:
## Types of changes

## Summary
Created transform commands for ssp filter by profile, and generate resolved profile catalog.

Also added regenerate option for ssp filter and for ssp assemble, so that ssp assemble will induce no changes on multiple runs if regenerate is False.

New transforms can be added based on these initial examples.

Added more functionality for loading models directly from the trestle directory.

Added test utility to extract the top level uuid of a model.

closes #854  
closes #811
## Key links:


trestle/core/commands/transform_oscal.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def transform(
+        self, trestle_root: pathlib.Path, model_type: str, model_name: str, output_name: str, profile_name: str
+    ) -> int:
+        """Transform the input file based on the profile and infer the output type."""
+        if model_type == const.MODEL_TYPE_SSP:
+            ssp_filter = SSPFilter()
+            return ssp_filter.filter_ssp(trestle_root, model_name, profile_name, output_name)
+        logger.warning(f'Transform operation not available for model type {model_type}')
+        return 1
```
butler54, line 55: The biggest problem I have with this approach is you have presumed if you get an ssp you automatically do a filter. 
I believe the approach should be that the transforms are 'named' 

e.g. `trestle transform --input input_ssp --output output_ssp --transform filter-by-profile -p my_profile`

because you could have:
`trestle transform --input input_ssp --output output_ssp --transform filter-by-components -c component -c component_b

that way we can address multiple types of transforms for a given stage.

butler54 (changes requested):
See feedback on interface.

trestle/core/commands/transform_oscal.py
fsuits: Redid it all more in line with your original plan.

tests/trestle/core/commands/transform_oscal_test.py
```diff
@@ -0,0 +1,136 @@
+# Copyright (c) 2020 IBM Corp. All rights reserved.
```
enikonovad, line 1: 2021

trestle/core/commands/author/ssp.py
```diff
@@ -123,46 +126,91 @@ def _init_arguments(self) -> None:
...
+
+    def _imp_reqs_equivalent(
+        self, imp_reqs: List[ossp.ImplementedRequirement], orig_imp_reqs: List[ossp.ImplementedRequirement]
+    ) -> bool:
+        """
+        Check if imp_reqs are the same except for internal uuids.
+
+        Create copy of each imp_req and set the uuids within it to match the original - then check equality.
+        """
+        if len(as_list(imp_reqs)) == len(as_list(orig_imp_reqs)):
```
enikonovad, line 139: I guess this is done as those two lists can be None, shall we mark them as optional in function's argument definition then?

trestle/core/commands/author/ssp.py
```diff
@@ -123,46 +126,91 @@ def _init_arguments(self) -> None:
...
+            for reqs in zip(imp_reqs, orig_imp_reqs):
```
enikonovad, line 140: If lists are None as expected above this will fail regardless, may be better to check if lists are none at the very beginning?

trestle/core/commands/author/ssp.py
```diff
@@ -123,46 +126,91 @@ def _init_arguments(self) -> None:
...
+        return True
 
     def _run(self, args: argparse.Namespace) -> int:
         log.set_log_level_from_args(args)
+        trestle_root = pathlib.Path(args.trestle_root)
 
+        # load the fresh imp_reqs from the markdown
         try:
+            md_path = trestle_root / args.markdown
+        except Exception as e:
```
enikonovad, line 165: I guess this is because `args.markdown` can be None? As it terminates command execution anyways, wouldn't it be easier to check if the argument is passed and notify the user if it is not?

trestle/core/commands/transform_oscal.py
```diff
@@ -0,0 +1,93 @@
...
+        model_type: str,
+        transform: str,
+        input_name: str,
+        transform_by: str,
+        output_name: str,
+        regenerate: bool
+    ) -> int:
+        """Transform the input file based on the profile and infer the output type."""
+        if model_type == const.MODEL_TYPE_SSP:
+            if transform == const.FILTER_BY_PROFILE:
```
enikonovad, line 79: Could be checked in one if statement

trestle/core/commands/transform_oscal.py
```diff
@@ -0,0 +1,93 @@
...
+                if not transform_by:
+                    logger.warning('A profile name must be provided for transform-by in the command.')
+                    return 1
+                ssp_filter = SSPFilter()
+                return ssp_filter.filter_ssp(trestle_root, input_name, transform_by, output_name, regenerate)
+        elif model_type == const.MODEL_TYPE_PROFILE:
+            if transform == const.GENERATE_RESOLVED_CATALOG:
```
enikonovad, line 86: Same as above

enikonovad (changes requested):
Small changes

tests/trestle/core/commands/transform_oscal_test.py
fsuits: fixed

trestle/core/commands/author/ssp.py
fsuits: I have confirmed they cannot be none so I removed as_list from each.

trestle/core/commands/author/ssp.py
fsuits: I have removed as_list.

trestle/core/commands/author/ssp.py
fsuits: This was a refactoring error.  I have restructured the try block to capture the overall assembly operation.

trestle/core/commands/transform_oscal.py
fsuits: This was intentional because I anticipate more options added here.  I have added a comment.

trestle/core/commands/transform_oscal.py
fsuits: ditto

trestle/core/commands/transform_oscal.py
enikonovad: Thought it could be the reason

enikonovad (approved):
LGTM

butler54 (approved):
LGTM

PR #876, opened 2021-11-18
feat: OSCO transformer support for OpenScap 1.3.5

### LINKED ISSUE #875: OSCO transformer update to handle OpenScap 1.35.

OSCO version 0.1.44, with OpenScap 1.3.5 has host name in different location than OSCO version 0.1.42, with OpenScap 1.3.3.

## Caveats / Assumptions

In 1.3.3 we get host name from \<target\>.
In 1.3.5 we need to get host name from \<target-facts\>\<fact name="urn:xccdf:fact:asset:identifier:host_name" type="string"\>

## Completion Criteria

OSCO transformer can handle both types of host name specification.

degenaro:
## Types of changes

## Summary

Support for OpenScap 1.3.5, in addition to 1.3.3.

Closes #875.

## Key links:

### DIFF
--- a/trestle/transforms/utils/osco_helper.py
+++ b/trestle/transforms/utils/osco_helper.py
@@ -61,7 +61,7 @@ t_results_map = Dict[str, Any]
 class RuleUse():
     """Represents one rule of OSCO data."""
 
-    def __init__(self, **args) -> None:
+    def __init__(self, args: Dict[str:str]) -> None:
         """Initialize given specified args."""
         self.id_ = args['id_']
         self.target = args['target']
@@ -83,10 +83,11 @@ class RuleUse():
         """Derive inventory key."""
         if self.host_name is None:
             # OpenScap 1.3.3
-            return self.target + ':' + self.target_type
+            rval = self.target + ':' + self.target_type
         else:
             # OpenScap 1.3.5
-            return self.host_name + ':' + self.target_type
+            rval = self.host_name + ':' + self.target_type
+        return rval
 
 
 class ComplianceOperatorReport():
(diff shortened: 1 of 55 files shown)


trestle/transforms/utils/osco_helper.py
```diff
@@ -61,36 +61,32 @@
...
-        scanner_name: str,
-        scanner_version: str,
-        idref: str,
-        version: str,
-        time: str,
-        result: str,
-        severity: str,
-        weight: str
-    ) -> None:
+    def __init__(self, **args) -> None:
```
butler54, line 64: I can see what you have done here to circumvent sonar complaining about too many arguments (I think it's limit is 7?).

However, this is pretty ugly as all the arguments are still required and there is now no longer an explicit spec. 
One idea would be to pass these as a dict (after all there will be no overhead as that is what the function call is doing anyway here).

TL;DR: It looks strange.

trestle/transforms/utils/osco_helper.py
```diff
@@ -61,36 +61,32 @@
...
+        self.severity = args['severity']
+        self.weight = args['weight']
+
+    @property
+    def inventory_key(self):
+        """Derive inventory key."""
+        if self.host_name is None:
+            # OpenScap 1.3.3
+            return self.target + ':' + self.target_type
+        else:
```
butler54, line 87: You don't need the else here as you have returned.

trestle/transforms/utils/osco_helper.py
```diff
@@ -217,19 +224,20 @@ def _parse_xml(self) -> Iterator[RuleUse]:
...
-                    benchmark_id,
-                    scanner_name,
-                    scanner_version,
-                    idref,
-                    version,
-                    time,
-                    result,
-                    severity,
-                    weight
+                    id_=id_,
```
butler54, line 227: See the comments above. I think a dict might actually make it clearer as the assignment and retrieval can happen at once.

butler54 (changes requested):
See the changes suggested. I'm happy to discuss as they do not change the logic.
One additional point:
It would be great if we can move away from doing filecmp with target files and instead can do attribute comparisons etc. If I change the json output formatting slightly it breaks the test (I can live with what we have right now in the tests).

From a Logic POV looks good to me.

trestle/transforms/utils/osco_helper.py
degenaro: I think is is 12 or 13, I missed by 1. Fixed.

trestle/transforms/utils/osco_helper.py
degenaro: Fixed.

trestle/transforms/utils/osco_helper.py
degenaro: Fixed.

butler54 (approved):
LGTM

PR #895, opened 2021-11-25
feat: Enforce OSCAL version and notify user

### LINKED ISSUE #581: OSCAL version should be checked at low level while file is read
Import will validate models in memory after loading, but models based on the wrong version of OSCAL won't load in the first place and will give a stream of pydantic validation errors like:

`WARNING:root:89 validation errors for Catalog
metadata -> responsible-parties
  value is not a valid list (type=type_error.list)
groups -> 0 -> controls -> 0 -> params -> 1 -> select -> how-many
  value is not a valid enumeration member; permitted: 'one', 'one-or-more' (type=type_error.enum; enum_values=[<HowMany.one: 'one'>, <HowMany.one_or_more: 'one-or-more'>])
groups -> 0 -> controls -> 3 -> controls -> 3 -> params -> 1 -> select -> how-many
  value is not a valid enumeration member; permitted: 'one', 'one-or-more' (type=type_error.enum; enum_values=[<HowMany.one: 'one'>, <HowMany.one_or_more: 'one-or-more'>])`

To prevent this from happening and to provide a user with a clear statement that the oscal version is not compatible, the OSCAL version should be checked early and while reading the file.

enikonovad:

## Types of changes

## Summary
Enforcing OSCAL files to have a OSCAL version `1.0.0`. When the wrong version is provided the error displayed to user will be:
```
trestle.core.commands.import_:82 WARNING: Error importing file: 4 validation errors for Catalog
metadata -> oscal-version -> __root__
  OSCAL version: 1.0.0-rc2 is not supported, use 1.0.0 instead. (type=value_error)
metadata -> responsible-parties
  value is not a valid list (type=type_error.list)
metadata -> a_wrong_field
  extra fields not permitted (type=value_error.extra)
metadata -> wrong_field_after
  extra fields not permitted (type=value_error.extra)

```

## Key links:

# Before you merge

- Check sonar

Closes #581


butler54:
I think that we probably should do a filter for when there is a version mis-match. If I try with a mis-matched catalog it will be a very large error messages.

My thoughts would be:
1. if we can idenify an version mismatch - state only that in the warning/error log and bury the full pydantic error in debug.
2. If we can't identify the version mismatch - we just log the error.

This would be 'fail safe', however, still a little brittle.

enikonovad:
Added a wrapper around `BaseModel` to print only `OSCAL version` error if present.

Now errors will look like this: 
if version is incorrect:
```
trestle.core.commands.import_:79 WARNING: Error importing file: OSCAL version: 1.0.0-rc2 is not supported, use 1.0.0 instead.
```

if version is correct, but body is not:
```
trestle.core.commands.import_:82 WARNING: Error importing file: 4 validation errors for Catalog
metadata -> responsible-parties
  value is not a valid list (type=type_error.list)
metadata -> a_wrong_field
  extra fields not permitted (type=value_error.extra)
metadata -> wrong_field_after
  extra fields not permitted (type=value_error.extra)
```

butler54 (changes requested):
The changes look good, however, collaborate with @fsuits to ensure that the changes to the OscalVersion class are injected via the scripts which optimise the classes.

Otherwise updates by nist will remove the code.

butler54 (approved):
LGTM with @fsuits changes

PR #990, opened 2022-01-12
feat: Validate OSCAL directories

### LINKED ISSUE #678: Refine handling of files extraneous files within oscal directories
Introducing spurious files within the oscal directory trees today leads to challenges with errors.

## Consistency and some liberty is needed in terms of where files can go:

/ANY FILE ALLOWED HERE (except conflicting with mandated directories such as catalogs.)
/catalogs/ <- Files allowed here
/catalogs/ <- Directories here will break trestle functionality but this is not easy to test for.
/catalogs/my_catalog/catalog.json
/catalogs/my_catalog/ <- README.md and other files allowed here.

/catalogs/my_catalog/catalog/metadata.json <-- no other files allowed here.
/catalogs/my_catalog/catalog/metadata

## Caveats:
- hidden files do not count
- Hidden files will be automatically deleted.

## Other issues:
-  File type inference is flakey - should we enforce single types throughout a model context (or project).
- Inconsistent recognition of consistent hidden files.

enikonovad:

## Types of changes

## Summary
Added a consistency check for OSCAL directories. No other files than `json | xml | yaml` or `md` are allowed to be in the directories. Hidden files are deleted automatically.

Also fixed a small bug related to `.keep` files on Windows

## Key links:

# Before you merge

Closes #678

### DIFF
--- a/trestle/core/commands/command_docs.py
+++ b/trestle/core/commands/command_docs.py
@@ -45,6 +45,7 @@ class CommandPlusDocs(CommandBase):
     """This class validates trestle-root argument.
 
     Trestle commands requiring trestle-root should extend from this class.
+    All commands that extend this class will validate the state of Trestle workspace.
     """
 
     def _validate_arguments(self, args):
@@ -52,6 +53,10 @@ class CommandPlusDocs(CommandBase):
         root = fs.get_trestle_project_root(args.trestle_root)
         if root is None:
             logger.error(f'Given directory {args.trestle_root} is not in a valid trestle root directory')
+            return CmdReturnCodes.TRESTLE_ROOT_ERROR.value
+        is_oscal_dir_valid = fs.check_oscal_directories(args.trestle_root)
+        if not is_oscal_dir_valid:
+            logger.error('OSCAL directories are not valid, please remove extra files specified above.')
             return CmdReturnCodes.TRESTLE_ROOT_ERROR.value
         args.trestle_root = root
         return CmdReturnCodes.SUCCESS.value
--- a/trestle/core/commands/init.py
+++ b/trestle/core/commands/init.py
@@ -16,6 +16,7 @@
 """Trestle Init Command."""
 import argparse
 import logging
+import os
 import pathlib
 import shutil
 from shutil import copyfile
@@ -27,6 +28,10 @@ import trestle.utils.log as log
 from trestle.core.commands.command_docs import CommandBase
 from trestle.core.commands.common.return_codes import CmdReturnCodes
 from trestle.core.err import TrestleError
+
+if os.name == 'nt':  # pragma: no cover
+    import win32api
+    import win32con
 
 logger = logging.getLogger(__name__)
 
@@ -70,7 +75,12 @@ class InitCmd(CommandBase):
             # Create directories
             for directory in directory_list:
                 directory.mkdir(parents=True, exist_ok=True)
-                file_path = pathlib.Path(directory) / const.TRESTLE_KEEP_FILE
+                if os.name == 'nt':
+                    file_path = pathlib.Path(directory) / const.TRESTLE_KEEP_FILE_WINDOWS
+                    atts = win32api.GetFileAttributes(str(file_path))
+                    win32api.SetFileAttributes(str(file_path), win32con.FILE_ATTRIBUTE_HIDDEN | atts)
+                else:
+                    file_path = pathlib.Path(directory) / const.TRESTLE_KEEP_FILE_LINUX
                 file_path.touch()
         except OSError as e:
             raise TrestleError(f'Error while creating directories: {e}')
--- a/trestle/core/commands/init.py
+++ b/trestle/core/commands/init.py
@@ -16,7 +16,6 @@
 """Trestle Init Command."""
 import argparse
 import logging
-import os
 import pathlib
 import shutil
 from shutil import copyfile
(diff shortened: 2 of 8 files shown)


trestle/core/commands/command_docs.py
```diff
@@ -53,5 +54,9 @@ def _validate_arguments(self, args):
         if root is None:
             logger.error(f'Given directory {args.trestle_root} is not in a valid trestle root directory')
             return CmdReturnCodes.TRESTLE_ROOT_ERROR.value
+        is_oscal_dir_valid = fs.check_oscal_directories(args.trestle_root)
+        if not is_oscal_dir_valid:
+            logger.error('OSCAL directories are not valid, please remove extra files specified above.')
+            return CmdReturnCodes.TRESTLE_ROOT_ERROR.value
```
fsuits, line 60: This is an example where it is a user error and not an error in trestle.  I think we want that to be a warning and errors are only for something wrong internal.  But I'm flexible.

trestle/core/commands/init.py
```diff
@@ -28,6 +29,10 @@
 from trestle.core.commands.common.return_codes import CmdReturnCodes
 from trestle.core.err import TrestleError
 
+if os.name == 'nt':  # pragma: no cover
+    import win32api
+    import win32con
```
fsuits, line 34: This may be ok as a check for windows if it is just for import.

trestle/utils/fs.py
```diff
@@ -61,6 +62,47 @@ def get_trestle_project_root(path: pathlib.Path) -> Optional[pathlib.Path]:
...
+    trestle_dir_walk = os.walk(root_path)
+    is_valid = True
+
+    for _, dirs, _ in trestle_dir_walk:
+        for d in dirs:
+            if d in MODEL_DIR_LIST:
+                is_valid = verify_trestle_folder(root_path / d)
+                if not is_valid:
+                    logger.error(
+                        f'OSCAL directory {root_path / d} contains unsupported files, '
```
fsuits, line 98: Does this work if someone has a *file* named catalogs in the main dir?

fsuits (changes requested):
Please look at the changes requested.  I think it would be good to clean things up at this point.  Overall changes look good - just need some refactor and more reuse.

trestle/core/commands/command_docs.py
enikonovad: I guess I wanted to emphasise to the user that this is an issue that needs to be fixed. The command will not proceed without this thus it is an error. But happy to make to log this at the INFO level if that is the convention we would like to follow.

trestle/utils/fs.py
enikonovad: As per discussion we allow any files and folders in the main directory, thus I only check the contents of OSCAL directories

trestle/core/commands/command_docs.py
fsuits: Normally I would do it as warning and not info - if the user made a mistake.  I use info for normal feedback the user needs - such as something being validated.  In this case given the need for the user to do something they may not be aware of - I think leaving it as error makes sense.

trestle/utils/fs.py
fsuits: This would definitely be a problem if there is a *file* named catalogs instead of a *directory* named catalogs.  I don't think we ever discussed this particular oddball case.  I'm just making sure that if allowed directories are present: catalogs, profiles, etc. - they really are directories and not files.  I wasn't clear what your code would do there.

enikonovad: I only iterate over directories in the main folder, I don't consider the files. We can always restrict users to not have files with such names as well

trestle/core/commands/command_docs.py
enikonovad: Changed it to warning

enikonovad:
> Is a non-trestle directory allowed at the top level?
> 
> degenaro:git$ cd shared-trestle-workspace degenaro:shared-trestle-workspace$ pwd /home/degenaro/git/shared-trestle-workspace degenaro:shared-trestle-workspace$ ls adjunct assessment-plans assessment-results catalogs component-definitions dist plan-of-action-and-milestones profiles system-security-plans target-definitions wild-side

Yes it is, anything is allowed in the main directory, the check only happens for OSCAL directories like `catalogs|profiles|etc`

trestle/core/commands/init.py
```diff
@@ -70,8 +75,14 @@ def _create_directories(self, root: pathlib.Path) -> None:
...
             for directory in directory_list:
                 directory.mkdir(parents=True, exist_ok=True)
-                file_path = pathlib.Path(directory) / const.TRESTLE_KEEP_FILE
-                file_path.touch()
+                if fs.is_windows():
+                    file_path = pathlib.Path(directory) / const.TRESTLE_KEEP_FILE_WINDOWS
+                    file_path.touch()
+                    atts = win32api.GetFileAttributes(str(file_path))
+                    win32api.SetFileAttributes(str(file_path), win32con.FILE_ATTRIBUTE_HIDDEN | atts)
+                else:
```
fsuits, line 83: Hmm - I just realized you wanted to use 'keep' on windows and not '.keep'.  It's actually fine to use .keep so I would just use the single const.TRESTLE_KEEP_FILE.  So all this code should reduce to simply `make_hidden_file(path/.keep)`. And it will do all the OS specific stuff.  The code should never have to call win32api outside of a few dedicated utility functions.

fsuits (approved):
LGTM

PR #999, opened 2022-01-21
fix: missing params in generated catalog and better hidden file handling on windows

fsuits:
## Types of changes

## Summary
catalog-generate was not inserting parameters in the markdown.  This now happens by default.
Windows now views all . files as hidden - in addition to all files with the hidden attribute set.  This is mainly because a clone will not set attributes - and they won't have been set in the first place from a unix system.
More code cleanup.
## Key links:

# Before you merge


tests/test_utils.py
```diff
@@ -431,3 +432,64 @@ def create_profile_in_trestle_dir(trestle_root: pathlib.Path, catalog_name: str,
...
+    """Check the total count of valid parts is the same, with recursion."""
+    n_a = get_total_valid_parts_count(a)
+    n_b = get_total_valid_parts_count(b)
+    if n_a != n_b:
+        logging.error(f'count of parts is different: {n_a} vs. {n_b}')
+        return False
+    return True
+
+
+def controls_equivalent(a: cat.Control, b: cat.Control) -> bool:
```
enikonovad, line 461: I have a strong feeling that comparison of the one instance to another should be placed within the instance's class. There is a `__eq__` method inside of the `BaseModel` that we can override for this purpose and place this code there. 
Of course, given that those files are autogenerated it might be tricky... I would like to know what do you think about it ?

tests/data/jinja/ssp_template.md.jinja
```diff
@@ -23,12 +23,12 @@ Test SSP
 
 {% for control in catalog_interface._control_dict.values() %}
 {% if control.group_id ==  group.id %}
-{# ### {{ catalog_interface.get_label(control.control) }}: {{ control.control.title}}
+{# ### {{ control_io_writer.get_label(control.control) }}: {{ control.control.title}}
```
enikonovad, line 26: We will also need to update the Jinja template in `cloud-ssp`, otherwise the release will break it, ill create an issue there.

enikonovad (approved):
LGTM - overall looks good, but I would much prefer if we could move the equality comparison to the class where its truly belongs.

tests/test_utils.py
fsuits: This is a relatively crude comparison and I wouldn't want people to think the controls are actually equal - they are just equivalent at a level needed for testing.  So I really think it should be confined to test_utils and only used for testing.

PR #1058, opened 2022-03-17
chore: Remove error duplicates and standarize exceptions part 2

### LINKED ISSUE #1009: Remove duplicate error/warning messages
Some errors trigger redundant messages, e.g.:

```
trestle.core.base_model:327 ERROR: Failed to parse OSCAL object: OSCAL version: 1.0.1 is not supported, use 1.0.0 instead.
trestle.core.commands.validate:52 ERROR: Error while validating contents of a trestle model: Error parsing file C:\tmp\test\profiles\fedramp\profile.json OSCAL version: 1.0.1 is not supported, use 1.0.0 instead.
```

This is due to an inner exception being caught and outputting a message, then later an outer exception is caught and this is repeated:

inner:
```python
        except Exception as e:
            logger.error(f'Failed to parse OSCAL object: {e}')
            raise err.TrestleError(f'Error parsing file {path} {str(e)}')
```
outer:
```python
        except TrestleError as e:
            logger.debug(traceback.format_exc())
            logger.error(f'Error while validating contents of a trestle model: {e}')
            return CmdReturnCodes.COMMAND_ERROR.value
```

## Caveats / Assumptions
The redundant info is distracting
## Completion Criteria
Decide how to handle this situation consistently
Also - should this be a warning?  It is due to bad data from the user and not a problem in trestle code.

enikonovad:
## Types of changes

## Summary
This is a second part of the code refactoring with the proposed structure to follow in the project when it comes to the exception handling and logging. The aim of the change is to remove the error duplicates and use appropriate logger level for logger messages.

The proposal consists of several parts:

### Logger level standardisation:
When to use which logging level:
- Info (logger.info): Provide information on successful execution. Confirm that things are working as expected.
- Warning (logger.warning): Issue a warning to the user that there is a problem that can lead to some issues or bugs. However, the problem does not halt the execution and operation proceeds.
- Error (logger.error): Indicates to a user that there is critical problem and execution cannot proceed. Such problem must end the execution and exit with appropriate code.

### Exception handling and logging
- `Logger.error` is no longer allowed anywhere but in the most outer level.
- `Logger.warning` and `Logger.info` allowed anywhere but the execution after `logger.warning` **must** proceed.
- Traceback will only be printed if command is executed in the verbose mode.
- Error exit codes are determined by exception handler in outer most level based on the exception type.
- Handling of inner exceptions - the inner exceptions should not be caught unless there is need to provide more information to the user on what went wrong. In this case, an appropriate instance of `TrestleError` should be raised.
  Example: 
```
except SomeException as e:
    raise TrestleError(f'Something went wrong during SSP generation: {e}')
```

- Each `_run()` function should be wrapped in `try-except` block.
- The outer most exception should be handled via `handle_generic_command_exception` with custom message and its output should be returned by `_run()`
- Raise exception rather than `return` command code manually.

## Key links:

# Before you merge

Closes #1009


trestle/core/commands/merge.py
```diff
@@ -146,37 +138,22 @@ def merge(cls, effective_cwd: Path, element_path: ElementPath, trestle_root: Pat
...
-        if target_model_path.exists():
-            logger.debug(
-                f'target model path does exist so target path is subdir with target alias {target_model_alias}'
-            )
-            target_model_path = target_model_path / target_model_alias
-        else:
-            logger.debug(f'target model filename does not exist so target path is target alias {target_model_alias}')
-            target_model_path = target_model_path / target_model_alias  # FIXME this is same as above
-        logger.debug(f'final target model path is {target_model_path}')
+        target_model_path = target_model_path / target_model_alias  # FIXME this is same as above
```
fsuits, line 144: Please see if that FIXME no longer applies or makes no sense.  If the code seems odd let me know and I want to check it before merge.  I can't see the context here.

trestle/core/commands/merge.py
```diff
@@ -146,37 +138,22 @@ def merge(cls, effective_cwd: Path, element_path: ElementPath, trestle_root: Pat
...
         # if target model is a file then handle file. If file doesn't exist, handle the directory,
         # but in this case it's a list or a dict collection type
         target_model_filename = target_model_path.with_suffix(file_ext)
         if target_model_filename.exists():
-            logger.debug(f'target model path with extension does exist so load distrib {target_model_filename}')
             _, _, target_model_object = ModelUtils.load_distributed(target_model_filename, trestle_root)
         else:
             target_model_filename = Path(target_model_path)
-            logger.debug(f'target model path plus extension does not exist so load distrib {target_model_filename}')
-            logger.debug(f'get collection type for model type {target_model_type}')
```
fsuits, line 172: I think it's fine to remove these debug lines but split and merge are hard to debug and it may be helpful to convert some to trace calls rather than remove them.

trestle/core/commands/partial_object_validate.py
```diff
@@ -60,40 +59,23 @@ def _run(self, args: argparse.Namespace) -> int:
             log.set_log_level_from_args(args)
             file_path: pathlib.Path = args.file.resolve()
             if not file_path.exists() or not file_path.is_file():
-                logger.error('File path provided does not exist or is a directory')
-                return CmdReturnCodes.COMMAND_ERROR.value
+                raise TrestleError('File path provided does not exist or is a directory')
+
             element_str: str = args.element
-            if ',' in element_str:
-                logger.error('Only a single element path is allowed.')
```
fsuits, line 67: is this check happening somewhere?  It should give warning.

trestle/core/commands/merge.py
enikonovad: Moved to trace instead

trestle/core/commands/partial_object_validate.py
enikonovad: Changed to warning

trestle/core/commands/merge.py
enikonovad: I revoked the changed for these lines, agree that all fixme should be checked. I will create an issue to go over all FIXMEs and see if they are relevant

fsuits (approved):
LGTM - this will be a good cleanup

PR #1062, opened 2022-03-21
fix: use control sort-id for sorting

### LINKED ISSUE #1018: Use sort-id for sorting controls
The current code uses the control name in some places for sorting controls, with an assumed structure to the name and associated numbering (i.e. it is not simply alphabetical by python's sorting routine).  But it would be better to use the control's sort-id, with fallback to the control name if none is available.

fsuits:
## Types of changes

## Summary
Now use sort-id in control props for sorting
The value is added to the yaml header if present in the control props

closes #1018
## Key links:

# Before you merge


trestle/core/control_io.py
```diff
@@ -71,6 +71,14 @@ def get_prop(part_control: Union[common.Part, cat.Control], prop_name: str) -> s
                 return prop.value.strip()
         return ''
 
+    @staticmethod
+    def get_sort_id(control: cat.Control, allow_none=False) -> str:
```
enikonovad, line 75: -> Optional[str]

trestle/core/control_io.py
```diff
@@ -1125,7 +1137,8 @@ def read_implemented_requirement(
 
         imp_req.statements = list(statement_map.values())
         ControlIOReader._insert_header_content(imp_req, header, control_id)
-        return imp_req
+        sort_id = header.get('sort-id', control_id)
```
enikonovad, line 1140: Question, if we have the `sort-id` in the control properties and the yaml header, which one will win? Is it also controlled by `override-header-values` flag ?

enikonovad (approved):
LGTM - please fix the typing

trestle/core/control_io.py
fsuits: fixed

trestle/core/control_io.py
fsuits: That function happens to be called only during ssp-assemble where no yaml header is provided.  So the sort-id only comes from the markdown header.  None of the assemble tools allow providing a yaml header, so it will always come from the markdown.

I changed write_control so that it explicitly handles sort-id according to overwrite-header-values in a consistent manner.  If it is already in the markdown control it will not get changed - but otherwise if it is not present it will be installed.

PR #1116, opened 2022-05-05
feat: validate refs and resources in catalogs and models

### LINKED ISSUE #675: Validator to ensure closed set of refs and resources
A resolved profile catalog should have a resource in back-matter for every reference in the document - and include no resources that are not referenced.  The references can be via uuid in a Link, or embedded as a link in a control's Part.Prose of the form `[foo](#uuidstring)`.

The generation of resolved catalogs will handle this, but a separate validator should be made to confirm that the links in general documents are similarly self-consistent in terms of refs in the document and resources in backmatter.

fsuits:
## Types of changes

## Summary
Added two validators.   One confirms all items refd by controls in catalog are present in resources
Other confirms that there are no duplicate param ids in catalog
Allowed validators to be quiet and suppress validation messages on import
improved handling of adding controls during catalog assembly
Updated docs

closes #675


trestle/core/catalog_interface.py
```diff
@@ -178,30 +182,44 @@ def get_sorted_controls_in_group(self, group_id: str) -> List[cat.Control]:
...
+        """Get all control ids from the control dict."""
         return self._control_dict.keys()
 
     def get_control(self, control_id: str) -> Optional[cat.Control]:
-        """Get control from catalog with this id using the dict."""
+        """Get control from the control dict with this id."""
         return None if control_id not in self._control_dict else self._control_dict[control_id].control
 
+    @staticmethod
+    def get_control_ids_from_catalog(catalog: cat.Catalog) -> List[str]:
```
enikonovad, line 202: I think adding a static function that takes catalog is a bit of overlap with the idea behind `CatalogInterface`. 
i.e.
What is the difference between this one and simply doing this:
```
cat_inter = CatalogInterface(catalog)
cat_inter.get_control_ids()
```

trestle/core/catalog_interface.py
```diff
@@ -362,38 +396,84 @@ def _update_all_controls_in_list(self, controls: List[cat.Control]) -> List[cat.
...
+                    node = found_group
+                else:
+                    raise TrestleError(f'No controls found in catalog for group {group.id}')
+            node.title = control_handle.group_title
+            node.class_ = control_handle.group_class
+        node.controls = as_list(node.controls)
+        try:
+            index = [control.id for control in as_list(node.controls)].index(control_handle.control.id)
+            del node.controls[index]
+        except ValueError:
```
enikonovad, line 445: I assume this is because of the potential non-existince of the value in the list?
If so, I think it is better to initialize the variable for the list and do the check if the element is in the list rather than trigger and catch the exception i.e.
```
control_ids = [control.id for control in as_list(node.controls)]
index = control_ids.index(control_handle.control.id) if control_handle.control.id in control_ids else None
if index:
   del node.controls[index]
```

trestle/core/catalog_interface.py
```diff
@@ -362,38 +396,84 @@ def _update_all_controls_in_list(self, controls: List[cat.Control]) -> List[cat.
...
+        """
+        # first go through the catalog and pull existing controls from the dict
+        for group in as_list(self._catalog.groups):
+            self._update_all_controls_in_group(group)
+
+        new_list = []
+        for control in as_list(self._catalog.controls):
+            new_control = self.get_control(control.id)
+            new_control.controls = self._update_all_controls_in_list(as_list(control.controls))
+            new_control.controls = none_if_empty(new_control.controls)
```
enikonovad, line 467: I wonder if that is something that can be done via pydantic during field initialisation. Potentially via [validators](https://pydantic-docs.helpmanual.io/usage/validators/) That would add consistency to the code, clean up all `none_if_empty` in the code, but will require to rewrite the generated models and add extra code there.

Not necessary change for this PR, but would be nice to have I think?

trestle/core/links_validator.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # find uuids in backmatter
+        links: List[str] = []
+        if model.back_matter and model.back_matter.resources:
+            links = [res.uuid for res in model.back_matter.resources]
+            seen: Set[str] = set()
+            dupes = [uuid for uuid in links if uuid in seen or seen.add(uuid)]
+            if dupes and not quiet:
+                logger.warning(f'Backmatter has  {len(dupes)} duplicate link uuids.')
+            logger.debug(f'Backmatter has {len(dupes)} duplicate link uuids: {dupes}')
```
enikonovad, line 51: I think this log should only be displayed if there are dupes?

enikonovad (changes requested):
Overall looks good, some small comments

trestle/core/catalog_interface.py
fsuits: This is part of the re-working that was done.  Some interface queries rely on information in the control dictionary - which comprises controls pulled out of the catalog into a separate standalone dictionary.  So for performance, queries against that dict are fast.  So an interface has added expense of needing to crate that dict.  But if I just want to know the controls in a given catalog, a one-pass recursive approach is best.  So it's a matter of doing a one-off scan of controls, vs. a series of queries that justifies the expense of building the dict.

trestle/core/catalog_interface.py
fsuits: From the web it seems like the valueerror approach is more standard and pythonic - and it involves only one linear pass as opposed to two - not that it really matters.  How about I make a new list function that uses your code so it's in one place, and I use it.

trestle/core/catalog_interface.py
fsuits: I'm not sure how that would work but happy to talk about it.

trestle/core/links_validator.py
fsuits: Oh - good catch thanks.

trestle/core/catalog_interface.py
enikonovad: Generally I think it is a bad practice to silence the exceptions, the two linear passes here are not too bad considering that we are dealing with relatively small lists in majority of the cases

trestle/core/catalog_interface.py
enikonovad: So would it be true to say that `cat_inter.get_control_ids() ==  CatalogInterface. get_control_ids_from_catalog(catalog)` ?

fsuits: The results are the same, but there is big expense in creating the control dict just for the purpose of getting the list of ids.  In my usage I need the control dict for one purpose - and I need to know the control ids in the actual catalog.  I have no need to instantiate a new catalog interface on top of the one I already have.  In addition I expect there will be other needs for a quick list of ids in a catalog - without other queries that warrant a cataloginterface.

enikonovad: okay, I just find the overhead of maintaining two separate functions with similar functionalities to be not ideal. But given that there is a purpose should be okay, let it stay this way

trestle/core/catalog_interface.py
fsuits: I created a new generic function in list_utils with no exceptions raised.  It takes a list, a value, and a lambda for getting the value of an element.

trestle/core/links_validator.py
fsuits: fixed

trestle/core/catalog_interface.py
enikonovad: I was hoping something like this would work (inside of the `Catalog`):
```
@validator("controls")
    def non_empty_controls(cls, values):
        values = values if values else None
        return values
```

But after some experimentation it seems like reassignment like `catalog.controls = new_controls` does not trigger the validator again. So this solution will not work. There could be some elegant solution in Pydantic for this, but it is fine for now

enikonovad (approved):
LGTM - thanks for the changes

PR #1137, opened 2022-06-08
fix: allow catalog groups with no id and fix validation of links and reference matching

fsuits:
## Types of changes

## Summary
Catalog group id's are optional so now create ids for any groups missing id's on import and give warning
Fixed bug counting links and references during validation
Added new load_validate.py file to validate files during load - which will also create group id's if needed
Then replaced all functions like load_top_level... to use the load_validate instead.

## Key links:

# Before you merge

### DIFF
--- a/trestle/core/commands/import_.py
+++ b/trestle/core/commands/import_.py
@@ -86,9 +86,8 @@ class ImportCmd(CommandPlusDocs):
                                                                 ).resolve()
 
             if desired_model_path.exists():
-                raise TrestleError(
-                    f'Cannot import because file to be imported here: {desired_model_path} already exists.'
-                )
+                logger.warning(f'Cannot import because file to be imported here: {desired_model_path} already exists.')
+                return CmdReturnCodes.COMMAND_ERROR.value
 
             if args.regenerate:
                 logger.debug(f'regenerating uuids in imported file {input_uri}')
@@ -121,7 +120,7 @@ class ImportCmd(CommandPlusDocs):
                     logger.warning(f'Validation of imported file {desired_model_path} did not pass.  Stopping import.')
                     rollback = True
             except TrestleError as err:
-                logger.warning(f'Import of {str(input_uri)} failed with validation error: {err}')
+                logger.error(f'Import of {str(input_uri)} failed with validation error: {err}')
                 rollback = True
 
             if rollback:
--- a/trestle/core/commands/import_.py
+++ b/trestle/core/commands/import_.py
@@ -29,6 +29,8 @@ from trestle.core.models.elements import Element
 from trestle.core.models.file_content_type import FileContentType
 from trestle.core.models.plans import Plan
 from trestle.core.remote import cache
+from trestle.core.validator import Validator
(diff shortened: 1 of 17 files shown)


enikonovad (commented):
I was playing around and got a bit confused with the use case. Say we have an existing catalog `ABC_catalog` that contains something like this (notice no group id):
```
"groups": [
      {
        "title": "",
        "controls": [
          {
            "id": "abc-1",
            "title": "ABC Test",
            "parts": [
              {
                "id": "abc-1_smt",
                "name": "statement",
                "prose": "Some Prose."
              }
            ]
          }
        ]
      }
   ]
```
And it is markdown representation.
Now we run `trestle author catalog-assemble -v --set-parameters --markdown md_catalogs/ABC_catalog --output ABC_catalog` for this catalog. 
It will result in error since control ids do not match the group.

My question is, is this a possible scenario? How can we expect the control ids to match group id?

fsuits:
OK I had to make many changes, but to summarize I rely on validation to create group id's as needed - but some of the load functions were not doing validation.  I added a new file load_validate to do loads and validations without causing circular imports.  I then changed all functions like load_top_level_model to use these methods - but they are still present in the tests, but that should be ok.

enikonovad (approved):
LGTM - looks solid, one suggestion is to add a component test that would import a catalog with no id, generate markdown and assemble back. And test that group id is assigned on all stages. Otherwise looks good.


```

trestle import -f <path_to_catalog>/CATALOG_WITH_NO_GROUP_ID.json -o A_CATALOG
trestle author catalog-generate --name A_CATALOG --output md_A_CATALOG
trestle author catalog-assemble --markdown md_A_CATALOG --output A_CATALOG

```

fsuits:
Your additional request ended up being important.  It turns out that Import runs the validation *after* importing the file, which means the validation doesnt get to fix the file.  I needed to add an early validation of the model in memory immediately after loading it.  This will prevent saving the model if it is not valid - but it will allow fixing it in ways that trestle is able to do during validation.  Validation is then run again on the saved file - as a final check.  A more elaborate test has been added to do the full import, generate markdown, then assemble with a catalog missing group id.

trestle/core/commands/import_.py
```diff
@@ -66,6 +68,14 @@ def _run(self, args: argparse.Namespace) -> int:
 
             model_read, parent_alias = fetcher.get_oscal(True)
 
+            # validate the loaded model in memory before writing out
+            # this will do any needed fixes to the file, such as assign missing catalog group ids
+            args_validate = argparse.Namespace(mode=const.VAL_MODE_ALL)
+            validator: Validator = validator_factory.get(args_validate)
+            if not validator.model_is_valid(model_read, True):
+                logger.warning(f'Validation of file to be imported {input_uri} did not pass.  Import failed.')
+                return CmdReturnCodes.COMMAND_ERROR.value
```
enikonovad, line 77: Since this ends the execution it should be an error instead.

enikonovad (approved):
LGTM - please use error instead of warning to confirm to logging in the rest of the commands

PR #1142, opened 2022-06-23
fix: Correct typo in README

folksgl:
Fix typo in README


fsuits (approved):
Thanks!  There may be other examples of Australian/British spelling vs. American and we will try to be consistent.

Issue #1161, opened 2022-08-25
Add special handling for display-name and namespaces in properties


fsuits:
## Issue description / feature objectives
Currently set-parameters do not show properties or namespace in the markdown yaml header - but it is desirable to have the new "display-name" property visible and editable in the header.

In addition, when a new display name is added, an appropriate namespace should be assigned to it.

Furthermore, if any new property is added to a  control or part, a namespace should be assigned to it.
## Caveats / Assumptions
We want to provide the needed editing capability without making the yaml header too busy or confusing.

~~No namespace values will ever be listed for properties in the yaml header - and instead they will either be as set in the source .json, or they will be specified by two new command line options described below.~~

## Completion Criteria

1. Provide special handling of the display-name property in set-parameters by including it in the markdown yaml header along with name, values, etc.
2. The user may add or edit that display name
3. ~~Provide new command line option display-name-ns for profile-assemble to specify the namespace for *all* display-name properties in the control.  If that value is provided on the command line, all display-name properties will have that namespace in the assembled profile.~~
4. Also provide new command line option default-namespace to use as the namespace for any added property that doesn't already have a namespace specified.  ~~This only applies to profile-assemble.~~
5. If other -assemble commands should behave this way - please edit this list.

vikas-agarwal76:
As discussed, during profile generate we will specify a CLI option default-ns. This would be shown in YAML header as a non-editable field. Currently we are assuming that "display-name" props of set-parameters will use this default ns. If in future they are using a different namespace then we will add another CLI option to specify the namespace for set-parameter properties (as we don't want to display the ns with each set-parameter property to keep set-parameter YAML headers simple).

For other properties at the control and subpart level, if the ns value is same as default-ns then the namespace for that property will not be shown in YAML header, else YAML header will show the specific ns for that property. If user creates or edits a property and don't specify a namespace in YAML header then the default-ns will be used during profile generate.
During profile-generate the default-ns CLI option will be optional. If not specified then the one from Makdown file will be used, else If specified then it will override the default namespace from markdown for all properties for which a namespace has not been explicitly specified in the YAML header. Properties with explicit namespace specified in YAML header will always have that namespace (not the default namespace).

fsuits:
To summarize the desired behavior:

1. A new command line option default-namespace will be added to profile generate and assemble
[...]

PR #1162, opened 2022-08-26
feat: handle display name as property and initial handling of namespace option

fsuits:
## Types of changes

## Summary
Handle display name as special property
namespace handling is only initially sketched in and will be filled in later.

partial #1161 
## Key links:

# Before you merge

### DIFF
--- a/tests/trestle/core/commands/author/profile_test.py
+++ b/tests/trestle/core/commands/author/profile_test.py
@@ -193,6 +193,8 @@ def test_profile_generate_assemble(
     if set_parameters:
         assert set_params[0].param_id == 'ac-1_prm_1'
         assert set_params[0].values[0].__root__ == 'all personnel'
+        assert set_params[0].props[0].name == const.DISPLAY_NAME
+        assert set_params[0].props[0].value.startswith('Pretty')
         assert set_params[1].param_id == 'ac-1_prm_2'
         assert set_params[1].values[0].__root__ == 'Organization-level'
         assert set_params[1].values[1].__root__ == 'System-level'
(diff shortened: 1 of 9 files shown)


tests/trestle/core/commands/author/profile_test.py
```diff
@@ -356,7 +359,8 @@ def test_profile_failures(tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPatc
         required_sections='NeededExtra',
         regenerate=False,
         version=None,
-        allowed_sections=None
+        allowed_sections=None,
+        namespace=''
```
enikonovad, line 363: What about the test where namespace is set to something?

trestle/core/resolver/modify.py
```diff
@@ -361,6 +365,11 @@ def _modify_controls(self, catalog: cat.Catalog) -> cat.Catalog:
                     self._set_parameter_in_control_or_loose(set_param)
             alters = self._profile.modify.alters
 
+        # an add with no by-id applies to the control
+        # if position is starting it should appear immediately after title
+        # if position is ending it should appear at end
+        # if not specified it defaults to ending
+        # if no by-id then before is treated as starting and after is treated as ending
```
enikonovad, line 372: Shall this be formalised in the function docs?

trestle/core/commands/author/profile.py
```diff
@@ -319,6 +332,7 @@ def assemble_profile(
             version: Optional version for the assembled profile
             required_sections: Optional List of required sections in assembled profile, as comma-separated short names
             allowed_sections: Optional list of section short names that are allowed, as comma-separated short names
+            default_namespace: Optional namespace to be used for properties that don't have namespace specfied
```
enikonovad, line 335: I don't see how this namespace is used in the assemble ?

enikonovad (changes requested):
I am missing a part here, how does namespace is propagated to assembled profile? I think now it is ignored...
Please add the test to generate and assemble the profile with namespace other than `''`

tests/trestle/core/commands/author/profile_test.py
fsuits: I have updated the PR description to indicate namespace is not functional and is only initially sketched in.  There is an issue describing how namespaces should work, so the matter is tracked.  This pr only addresses new functionality for display-name.

trestle/core/resolver/modify.py
fsuits: this is a private function so I don't think it warrants full documentation, and the point of those comments is just a reminder for the logic of how position works, and how the defaults are specified by NIST.

trestle/core/commands/author/profile.py
fsuits: it isn't yet but it will be finished after demo

trestle/core/resolver/modify.py
enikonovad: The problem is that comments like that became outdated as code develops, easier to change docs rather than comment. But fine to keep for now

enikonovad (approved):
LGTM - approving as the work is partial. Please link this PR to the issue for the easier tracking in the future

PR #1198, opened 2022-10-10
feat: show inherited props in yaml header for profile-generate markdown

### LINKED ISSUE #1187: Handling chaining of profiles properly
When multiple profiles are chained one after another, the last profile's Markdown files should show the parameters, props, and sections properly to differentiate ones that are inherited from other profile and ones that are added in the current profile.

### Sections
Sections inherited from imported profiles/catalogs should be shown before the "Editable Content" header. Any changes in these sections should be ignored.

Only sections added after the "Editable Content" header will be part of this profile.

### Parameters
x-trestle-set-params:
  ra-5_prm_1:
    profile-values: weekly operating system/infrastructure and databases; weekly container
      images; quarterly web applications; annual source code/static analysis
    values:
    display-name: RA-5 (a)

"values" will always show the inherited value from imported profiles/catalog, whereas "profile-values" will show the value set in the current profile.

### Properties
Currently we have x-trestle-add-props: only which contain the properties . We need to divide the props in YAML header into 2 groups - one inherited from imported profiles/catalogs that are only for display purpose and other for props added in the current profile.

We can retain x-trestle-add-props: to show props added in current profile and have another header like - "x-trestle-inherited-props" to show props that are inherited and cannot be updated.

Keeping the x-trestle-add-props header for props added in current profile will simplify the migration from current Markdown as inherited props are currently not shown and will be shown in a new header.

fsuits:
## Types of changes

## Summary
New feature to list inherited props added by profiles in profile-generate
Refactored code with more comprehensions instead of loops
Added additional tests to check proper update of contents in resolved profile after updates to upstream catalog and controls
Clarified handling of sections to confirm CLI value of sections overrides what is in the yaml header

closes #1187 

## Key links:

# Before you merge


trestle/common/list_utils.py
```diff
@@ -79,7 +79,33 @@ def delete_item_from_list(item_list: List[Any], value: Any, key: Callable[[Any],
     return item_list
 
 
+def get_item_from_list(
+    item_list: Optional[List[Any]], value: Any, key: Callable[[Any], Any], remove: bool = False
+) -> Any:
+    """Get item from list if present."""
```
enikonovad, line 85: Please provide description for the function args as well

trestle/common/list_utils.py
```diff
@@ -79,7 +79,33 @@ def delete_item_from_list(item_list: List[Any], value: Any, key: Callable[[Any],
     return item_list
 
 
+def get_item_from_list(
+    item_list: Optional[List[Any]], value: Any, key: Callable[[Any], Any], remove: bool = False
+) -> Any:
```
enikonovad, line 84: Please use `TG`  instead of `Any` here. The problem with `Any` is that Python type checker will not validate the type at all, as _anything_ can be returned. If we instead use `TG` the types will be "linked" and type checker will work.

Consider a very simple identity function
```
def identity(a: Any) -> Any:
    return a
```

And call it 
```
a_number = identity(23)
print(a_number + "test") <--- This will not notify user of the type error.
```

However, if we use `TG` it will.

trestle/common/list_utils.py
```diff
@@ -79,7 +79,33 @@ def delete_item_from_list(item_list: List[Any], value: Any, key: Callable[[Any],
...
+    """Get item from list if present."""
+    if not item_list:
+        return None
+    keys = [key(item) for item in item_list]
+    item = None
+    if value in keys:
```
enikonovad, line 90: What if there are multiple values in the array? Should we return multiple values then? If not specify in the doc that only the first one will be returned

trestle/common/list_utils.py
```diff
@@ -79,7 +79,33 @@ def delete_item_from_list(item_list: List[Any], value: Any, key: Callable[[Any],
...
+    return get_item_from_list(item_list, value, key, True)
+
+
 def delete_list_from_list(item_list: List[Any], indices: List[int]) -> None:
     """Delete a list of items from a list based on indices."""
     for index in sorted(indices, reverse=True):
         del item_list[index]
+
+
+def merge_dicts(dest: Optional[Dict[str, str]], src: Optional[Dict[str, str]]) -> Dict[str, str]:
```
enikonovad, line 109: Should we follow a convention of bash commands and do src, dest instead of dest, src?

trestle/core/control_interface.py
```diff
@@ -518,10 +512,10 @@ def is_withdrawn(control: cat.Control) -> bool:
...
         This is determined by property with name 'status' with value 'Withdrawn'.
         """
-        for prop in as_list(control.props):
-            if prop.name and prop.value:
-                if prop.name.lower().strip() == 'status' and prop.value.lower().strip() == 'withdrawn':
-                    return True
+        for _ in as_filtered_list(
+                control.props,
+                lambda p: strip_lower_equals(p.name, 'status') and strip_lower_equals(p.value, 'withdrawn')):
+            return True
```
enikonovad, line 518: if as_filtered_list(.... -> return True?

trestle/core/control_context.py
```diff
@@ -53,6 +53,7 @@ class ControlContext:
     allowed_sections: Optional[str] = None
     comp_def: Optional[comp.ComponentDefinition] = None
     comp_name: Optional[str] = None
+    inherited_props: Optional[Dict[str, Any]] = None
```
enikonovad, line 56: What is Any here?

trestle/core/profile_resolver.py
```diff
@@ -29,17 +33,49 @@ class ProfileResolver():
...
+        cat_interface = CatalogInterface(res_cat)
+        for control in cat_interface.get_all_controls_from_dict():
+            part: com.Part = pop_item_from_list(control.parts, TRESTLE_INHERITED_PROPS, lambda p: p.name)
+            if part:
+                props_list: List[Dict[str, Any]] = []
+                for prop in as_list(part.props):
+                    props_list.append({'name': prop.name, 'value': prop.value})
+                for sub_part in as_list(part.parts):
+                    for prop in as_list(sub_part.props):
+                        props_list.append({'name': prop.name, 'value': prop.value, 'part_name': sub_part.title})
```
enikonovad, line 61: what about sub_part.part?

enikonovad (changes requested):
Nice to see more list comprehensions as it cleans up the code. Please see some comments regarding the typing

trestle/common/list_utils.py
fsuits: Done

trestle/common/list_utils.py
fsuits: Now using TG for the items and TG2 for the value and callable result

trestle/common/list_utils.py
fsuits: Docs now emphasize first

trestle/common/list_utils.py
fsuits: For me this is most natural and it's what I encounter more.  Either way can be confusing - but it is documented.

trestle/core/control_interface.py
fsuits: I think it's ok this way because it allows a short-circuit exit on the first hit.  I don't think just using `if` would avoid evaluating the entire array if there are multiple hits.

trestle/core/control_context.py
fsuits: The Any is itself a recursive Dict of [str, Dict[str, Dict[str]]]...] - so I think this is ok.

trestle/core/profile_resolver.py
fsuits: This is a structure that trestle creates and it is not recursive.  It's just a way to keep track of added props.  Props are added either at the control level as a single list, or by_id as a list bound a by_id sub_part.  It is described in the docs of that function.

trestle/core/control_interface.py
enikonovad: I meant just simply checking if the resulting array is empty or not would be the same?

trestle/core/profile_resolver.py
enikonovad: Seems a bit limiting to not display props in the deepest parts, but if that is by design then okay.

enikonovad (approved):
LGTM

trestle/core/profile_resolver.py
fsuits: All added props are captured in this list - they can be added at any level - by-id - and they will appear in this list.  There is no need for a deep list because this is just a list of added props that trestle uses to feed them out to the catalog writer.  And then the list is deleted from the control.

trestle/core/control_interface.py
fsuits: If you have 10 million items and the first is a hit then the loop i have will immediately exit and I don't think the `if` approach will do that.  I believe it will evaluate the entire list.  If that is incorrect it would be good to know.

PR #1288, opened 2023-02-14
feat: allow remote profiles to reference catalogs and profiles by relative path in href

### LINKED ISSUE #1287: Remote load of models needs better handling of imports that are local to a remote source
In general trestle can import a model specified by a remote url, but currently any links or references in that model that are local to itself will not be followed properly since they are assumed to be local to the trestle workspace.  None of this matters if all files are local to trestle or if all links are complete urls.

As a concrete example, if you load the remote nist profile it will load fine, but if its imports are simply `nist_sp_...json` instead of `https://nist-oscal/...` (i.e. a global url rather than a local file name) the corresponding linked file will not be found.

Solutions are either to require all paths to be explicit, or to assume simple file paths in a link are relative to the parent model.

fsuits:
## Types of changes

If you try to resolve a profile at a remote url, and its href to the catalog is local to its directory - it will fail.  This has now been updated so that it works.

closes #1287 

## Key links:

# Before you merge

### DIFF
--- a/tests/trestle/core/profile_resolver_test.py
+++ b/tests/trestle/core/profile_resolver_test.py
@@ -435,6 +435,6 @@ def test_profile_resolver_no_params(tmp_trestle_dir: pathlib.Path) -> None:
 
 def test_remote_profile_relative_cat(tmp_trestle_dir: pathlib.Path) -> None:
     """Test profile resolver with remote profile and import of relative catalog path."""
-    profile_path = 'https://raw.githubusercontent.com/usnistgov/oscal-content/main/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_LOW-baseline_profile.json'  # noqa E501
+    profile_path = 'https://raw.githubusercontent.com/usnistgov/oscal-content/feature-basic-end-to-end-example/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_LOW-baseline_profile.json'  # noqa E501
     resolved_cat = ProfileResolver.get_resolved_profile_catalog(tmp_trestle_dir, profile_path)
     assert len(resolved_cat.groups) > 10
--- a/tests/trestle/core/profile_resolver_test.py
+++ b/tests/trestle/core/profile_resolver_test.py
@@ -431,3 +431,10 @@ def test_profile_resolver_no_params(tmp_trestle_dir: pathlib.Path) -> None:
     catalog_str = catalog.oscal_serialize_json()
     # make sure no moustaches remain that would confuse jinja
     assert '{{' not in catalog_str
+
+
+def test_remote_profile_relative_cat(tmp_trestle_dir: pathlib.Path) -> None:
+    """Test profile resolver with remote profile and import of relative catalog path."""
+    profile_path = 'https://raw.githubusercontent.com/usnistgov/oscal-content/main/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_LOW-baseline_profile.json'  # noqa E501
+    resolved_cat = ProfileResolver.get_resolved_profile_catalog(tmp_trestle_dir, profile_path)
+    assert len(resolved_cat.groups) > 10
--- a/trestle/core/remote/cache.py
+++ b/trestle/core/remote/cache.py
@@ -100,7 +100,9 @@ class FetcherBase(ABC):
                 self._do_fetch()
                 return True
             except Exception as e:
-                raise TrestleError(f'Cache update failure for {self._uri}: {e}.') from e
+                raise TrestleError(
+                    f'Cache update failure for {self._uri}.  Please confirm the file is json and not html: {e}.'
+                ) from e  # noqa E501
         return False
 
     def get_raw(self, force_update=False) -> Dict[str, Any]:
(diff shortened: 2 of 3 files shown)


tests/trestle/core/profile_resolver_test.py
```diff
@@ -431,3 +431,10 @@ def test_profile_resolver_no_params(tmp_trestle_dir: pathlib.Path) -> None:
     catalog_str = catalog.oscal_serialize_json()
     # make sure no moustaches remain that would confuse jinja
     assert '{{' not in catalog_str
+
+
+def test_remote_profile_relative_cat(tmp_trestle_dir: pathlib.Path) -> None:
+    """Test profile resolver with remote profile and import of relative catalog path."""
+    profile_path = 'https://raw.githubusercontent.com/usnistgov/oscal-content/main/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_LOW-baseline_profile.json'  # noqa E501
+    resolved_cat = ProfileResolver.get_resolved_profile_catalog(tmp_trestle_dir, profile_path)
```
enikonovad, line 439: This file can potentially get deleted, modified etc. Lets fix it to a particular commit/tag instead (the URL should point to that commit/tag instead).

trestle/core/remote/cache.py
```diff
@@ -278,6 +280,8 @@ def _do_fetch(self) -> None:
                     raise TrestleError(f'Cache update failure with bad inputenv var: {err_str}')
         if self._username is not None and self._password is not None:
             auth = HTTPBasicAuth(self._username, self._password)
+
+        # attempt get of the file from url
```
enikonovad, line 284: Comment is not necessary here

trestle/core/resolver/_import.py
```diff
@@ -81,6 +88,19 @@ def __init__(
...
+        if uri_type == cache.FetcherFactory.UriType.LOCAL_FILE and self._parent_url_root:
+            self._import.href = self._parent_url_root + self._import.href
+        # if href is now a remote path, capture its parent path for possible use with child imports that are relative
+        if uri_type not in [cache.FetcherFactory.UriType.LOCAL_FILE, cache.FetcherFactory.UriType.TRESTLE]:
+            head_path, _ = os.path.split(self._import.href)
+            if head_path and head_path[-1] != '/':
+                head_path += '/'
+            self._parent_url_root = head_path
+            logger.info('parent head path %s', head_path)
+        logger.info('import href is %s', self._import.href)
```
enikonovad, line 102: No need to inform user on it, put it to debug level

trestle/core/resolver/_import.py
```diff
@@ -81,6 +88,19 @@ def __init__(
                     f'Back matter resource resolution needed for profile import failed with error: {str(e)}'
                 )
 
+        uri_type = cache.FetcherFactory.get_uri_type(self._import.href)
+        # if this looks like a relative path to remote source, append parent path
+        if uri_type == cache.FetcherFactory.UriType.LOCAL_FILE and self._parent_url_root:
+            self._import.href = self._parent_url_root + self._import.href
+        # if href is now a remote path, capture its parent path for possible use with child imports that are relative
+        if uri_type not in [cache.FetcherFactory.UriType.LOCAL_FILE, cache.FetcherFactory.UriType.TRESTLE]:
```
enikonovad, line 96: Just in case we put other Uri types in the future, it is better to say for which cases it does apply rather than the opposite as it will be incorrect

trestle/core/resolver/_import.py
```diff
@@ -81,6 +88,19 @@ def __init__(
...
+            head_path, _ = os.path.split(self._import.href)
+            if head_path and head_path[-1] != '/':
+                head_path += '/'
```
enikonovad, line 99: Use `os.path.dirname(self._import.href)` and then append two paths so you don't have to manually add `\` at the end

enikonovad (changes requested):
Please see the comments

trestle/core/resolver/_import.py
fsuits: thanks - I had meant for them to be debug

trestle/core/resolver/_import.py
fsuits: That would of course also fail if the new types were then missed.  I have made an explicit function defined right next to the enums so it should be more manageable and abstracted.

trestle/core/resolver/_import.py
fsuits: I have been burned by trusting some of these path functions, including .stem, that don't work as expected.  So when it is very simple and clear I prefer an explicit operation.  But I will make the change to use os here.

trestle/core/remote/cache.py
fsuits: removed - but it's there because previously it appeared linked to the password query, while this get() is generic.

tests/trestle/core/profile_resolver_test.py
fsuits: There is one tag available but it does not work due to its href loading xml.  I specified a branch that does work.  Either way - it will need to be updated as oscal evolves.

enikonovad (approved):
LGTM

Issue #1282, opened 2023-02-09
Reduce the number of JSON test data in favour of generating it on the fly


enikonovad:
## Issue description / feature objectives
Currently Trestle tests use a lot of hardcoded test data which is hard to maintain and update. Instead, we should either generate it on the fly from the schema or have a few common test files and modify their content in-memory to test specific cases.

This will allow to reduce the load on the developers when updating to the new OSCAL schemes or writing new tests.

## Completion Criteria
1. Review current JSON test data and access its need.
2. Explore the ways to reduce the number of test data, whether by generating it on the fly or modifying a "base" files as desired.
3. Provide a new functionality for the tests to fetch or generate the required data. 
4. Test's logic should remain the same.

PR #1298, opened 2023-02-17
feat: Add ability to view version of the individual OSCAL object

### LINKED ISSUE #1292: Add ability to view version of the particular OSCAL object
As user, I would like to be able to get the version of the OSCAL object that I work with without opening the file.
Such functionality will be needed for automation as well.

The task is to extend `trestle version` functionality to take a specific OSCAL model and return its metadata version.

## Completion Criteria
`trestle version -n <object_name> -t <type>` should return metadata version of that object.
`trestle version` logic should remain the same and return Trestle version and used OSCAL version.

enikonovad:
## Types of changes

## Summary

Added ability to view the metadata version of the individual OSCAL objects.

`trestle version -n <object_name> -t <type>` will return metadata version of that object.
`trestle version` has the same logic as previously

## Key links:

# Before you merge

Closes #1292


fsuits:
The validate command also allows model name and model type, but it finds the model using `get_models_of_type` in `model_utils`.  See line 62 in `validator.py`  This would allow support for all (and future) model types, but it does require spelling out the full model type.  See the `-h` help command for `validate`.

We could use some kind of lookup based on short names like `ssp` but if so the validate command should be similarly changed.

So I would avoid using the term 'folder' and just refer to model name and model type - with the full type.

For test coverage you can just generate models on the fly, set the version or whatever as you like, and save them out - then query the version.  This could even be done in a loop over all model types, generate random strings for version et al., and not rely on existing test data.

If there is a reason you only want to support a subset of types then the above may not apply, but I would aim to keep things consistent with `validate`.

enikonovad:
@fsuits , good suggestions will address it.

enikonovad:
@fsuits I have addressed the comment and now the `trestle version` will support all models similarly to `validate`.
Also did a small refactoring (mostly renaming) of existing functionality in `model_utils`. This class will need further revision.

As for the tests, I think we can postpone the generation of the models on the fly until the #1282 is completed

trestle/core/commands/version.py
```diff
@@ -30,7 +33,44 @@ class VersionCmd(CommandBase):
...
 
+    def _init_arguments(self) -> None:
+        self.add_argument('-n', '--name', help='Name of the OSCAL object (folder name)', default=None, type=str)
+        self.add_argument(
+            '-t',
+            '--type',
+            help='Name of the OSCAL type (catalog, profile, component-definition, system-security-plan, etc)',
+            default=None,
+            type=str
+        )
```
fsuits, line 44: I would say "Name of the OSCAL model" (and not mention folder name).  'Object' sounds lower level than model - and this only deals with what we refer to as OSCAL models.  None of the CLI commands mentions folder name, so this would stand out in that regard.  Models have a type and they are stored in trestle somehow by name - so just refer to the model name and type.

And on line 41 I would just say "Type of the model being queried: (catalog ...)"  This emphasizes the -t as type, and I don't think we need to say "name of the type".

fsuits (approved):
LGTM with minor comment

trestle/core/commands/version.py
enikonovad: done

PR #1299, opened 2023-02-21
fix: allow edit of rule param values during component assemble

### LINKED ISSUE #1297: Add test to exercise setting of parameter value during component-assemble and update docs
~~Currently there is no test to confirm the ability to set a value for a parameter via markdown during component assemble.  Also the docs are not clear on how to do this.  The current code expects new value entries with the name "ssp-value" for new values to be assigned in the component.  This check happens at line 570 of `control_reader.py`~~

~~To be consistent with `profile-assemble` using `profile-value` and `ssp-assemble` using `ssp-value` we should instead use here `component-value`.  The code at line 570 only executes during component assemble, so it is ok for it to be specific to that.~~

~~See `profile-assemble` and `ssp-assemble` for similar examples of setting parameters via markdown.~~ 

~~The current component_generate test should be extended to edit the markdown and set a new value prior to assemble - and confirm the new setparameter has been added or updated.  The docs in `ssp_profile_catalog_authoring.md` should also be updated with an example showing how to update the values via markdown.~~

On discussion with @vikas-agarwal76 it may be better to leave this as read-only for now, so the task is simplified to change the docs to make it clear the control parameter values are *not* settable - and to clarify the code so there is nothing suggesting the values will be assembled into the json model.

fsuits:
## Types of changes

## Summary
The code was not allowing edit of rule parameter values during component-generate and component-assemble.  This is now working properly and has a corresponding test.

closes #1297
## Key links:

# Before you merge

### DIFF
--- a/scripts/gen_oscal.py
+++ b/scripts/gen_oscal.py
@@ -78,7 +78,14 @@ def generate_model(full_name, out_full_name):
 
 
 def generate_models():
-    """Generate all models including 3rd party."""
+    """
+    Generate all models including 3rd party.
+
+    IMPORTANT NOTE: This script will leave the temporary classes in oscal/tmp for reference, but they should be deleted
+    when no longer needed because some of the operations in the build will seek all .py files, and will try to act
+    on these temporary files - giving an error.  So be sure to delete the oscal/tmp directory when done generating
+    new classes.
+    """
     logger.info('generating models')
     out_dir = Path('trestle/oscal')
     out_dir.mkdir(exist_ok=True, parents=True)
--- a/tests/trestle/core/commands/author/component_test.py
+++ b/tests/trestle/core/commands/author/component_test.py
@@ -23,7 +23,7 @@ from tests import test_utils
 
 import trestle.core.generic_oscal as generic
 import trestle.oscal.component as comp
-from trestle.common import const, model_utils
+from trestle.common import const, file_utils, model_utils
 from trestle.core.commands.common.return_codes import CmdReturnCodes
 from trestle.core.markdown.markdown_processor import MarkdownProcessor
 
@@ -101,8 +101,8 @@ def test_component_generate(tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPa
 
     file_checker = test_utils.FileChecker(tmp_trestle_dir / md_path)
 
-    generate_cmd = f'trestle author component-generate -n {comp_name} -o {md_path}'
     # confirm it overwrites existing md identically
+    generate_cmd = f'trestle author component-generate -n {comp_name} -o {md_path}'
     test_utils.execute_command_and_assert(generate_cmd, CmdReturnCodes.SUCCESS.value, monkeypatch)
 
     # all files should be the same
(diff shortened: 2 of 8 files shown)


scripts/gen_oscal.py
```diff
@@ -78,7 +78,14 @@ def generate_model(full_name, out_full_name):
...
 def generate_models():
-    """Generate all models including 3rd party."""
+    """
+    Generate all models including 3rd party.
+
+    IMPORTANT NOTE: This script will leave the temporary classes in oscal/tmp for reference, but they should be deleted
+    when no longer needed because some of the operations in the build will seek all .py files, and will try to act
+    on these temporary files - giving an error.  So be sure to delete the oscal/tmp directory when done generating
+    new classes.
+    """
```
enikonovad, line 88: That sounds like a good issue to address. Shall we create an issue to remove tmp folder automatically?

tests/trestle/core/commands/author/component_test.py
```diff
@@ -119,10 +119,23 @@ def test_component_generate(tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPa
     creation_time = assem_comp_path.stat().st_mtime
     assert model_utils.ModelUtils.models_are_equivalent(orig_component, assem_component, True)
 
+    # assemble again and confirm it is not written out since no change
     test_utils.execute_command_and_assert(assemble_cmd, CmdReturnCodes.SUCCESS.value, monkeypatch)
     assert creation_time == assem_comp_path.stat().st_mtime
 
-    ac1_path = tmp_trestle_dir / 'md_comp/comp_aa/comp_prof_aa/ac/ac-1.md'
+    # edit a rule param value
+    new_text = '      component-values:\n        - inserted value\n'
```
enikonovad, line 127: could we also test setting multiple values?

trestle/common/const.py
```diff
@@ -508,6 +508,18 @@
   #
 """
 
+YAML_RULE_PARAM_VALUES_COMPONENT_COMMENT = """  # You may set new values for rule parameters by adding
+  #
+  # component-values:
+  #   - value 1
+  #   - value 2
+  #
+  # below a section of values:
```
enikonovad, line 517: How does the values relate to the param-id ? In other words, how can I set a specific set-parameter in component-definition?

trestle/core/control_interface.py
```diff
@@ -1091,14 +1091,33 @@ def insert_imp_req_into_component(
...
                 if imp_req.control_id == new_imp_req.control_id:
+                    _, imp_req_param_dict, _ = ControlInterface.get_rules_and_params_dict_from_item(imp_req)
+                    imp_req_rule_param_ids = [d['name'] for d in imp_req_param_dict]
                     status = ControlInterface.get_status_from_props(new_imp_req)
                     ControlInterface.insert_status_in_props(imp_req, status)
                     imp_req.description = new_imp_req.description
                     statement_dict = {stat.statement_id: stat for stat in as_list(imp_req.statements)}
+                    # update set parameter values with values from markdown - but only for rule param vals
+                    for set_param in as_list(new_imp_req.set_parameters):
+                        if set_param.param_id in (control_imp_rule_param_ids + imp_req_rule_param_ids):
```
enikonovad, line 1108: rather a stylistic comment but Sonar would agree, we can use "guard clauses" to make code cleaner. Instead of doing something when condition is met, we can skip the step if condition is not met
```
if set_param.param_id not in ...
    continue
    
# Code goes here
```

enikonovad (approved):
LGTM - just a few curious questions/minor suggestions above

scripts/gen_oscal.py
fsuits: I think it's important for new people exposed to the script to have the tmp files retained so they can see them and understand the subsequent processing.  So I don't know of a way to remove them automatically.  The generation of the classes is inherently hands-on and has manual elements because we have no idea what future releases of OSCAL will look like.

tests/trestle/core/commands/author/component_test.py
fsuits: I think this would be a good learning task for Alejandro.  Can you sketch what is involved to him?  Otherwise I can do it Thurs.

trestle/common/const.py
fsuits: I think it will be clear to users that each "values:" list is bound to a single parameter id, and the new "component-values:" goes directly below it.  If you see the full markdown header with the comment it should make sense.

trestle/core/control_interface.py
fsuits: This could be another thing to mention to Alejandro - and also convey how you review code and provide feedback.  I can address it Thurs if he is not able to.

trestle/common/const.py
enikonovad: I see, that make sense then. Setting multiple values for the different parameters could be a good test case as well

AleJo2995:
Hey @fsuits and @enikonovad I want to propose to take ownership of the changes needed and solve them when needed accordingly and the review them with both of you. I will be commenting, asking and pushing changes in the branch where it makes sense to do it so, as well as adding review for the general code.

PR #1305, opened 2023-02-24
fix: Give warnings when component references control not loaded by profile for comp-gen and ssp-gen

fsuits:
## Types of changes

## Summary
Now give warnings if component references control not loaded by profile during ssp-gen and comp-gen
## Key links:

# Before you merge

### DIFF
--- a/trestle/common/log.py
+++ b/trestle/common/log.py
@@ -50,8 +50,7 @@ def set_global_logging_levels(level: int = logging.INFO) -> None:
     """
     # This line stops default root loggers setup for a python context from logging extra messages.
     # DO NOT USE THIS COMMAND directly from an SDK. Handle logs levels based on your own application
-    # set propagate to True in order to capture warning messages in tests via caplog
-    _logger.propagate = True
+    _logger.propagate = False
     # Remove handlers
     _logger.handlers.clear()
     # set global level
--- a/trestle/common/log.py
+++ b/trestle/common/log.py
@@ -50,7 +50,8 @@ def set_global_logging_levels(level: int = logging.INFO) -> None:
     """
     # This line stops default root loggers setup for a python context from logging extra messages.
     # DO NOT USE THIS COMMAND directly from an SDK. Handle logs levels based on your own application
-    _logger.propagate = False
+    # set propagate to True in order to capture warning messages in tests via caplog
+    _logger.propagate = True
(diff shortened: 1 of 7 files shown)


trestle/core/control_interface.py
```diff
@@ -1109,13 +1109,23 @@ def insert_imp_req_into_component(
...
+                        continue
+                    # but first check if the parameter was already set with the same value in the control_imp
+                    # if so we don't need to insert a new set_param in imp_req
+                    for dest_param in as_list(control_imp.set_parameters):
+                        if dest_param.param_id != set_param.param_id:
+                            continue
+                        if dest_param.values == set_param.values:
+                            found = True
+                            break
                     if found:
```
enikonovad, line 1129: Hmm the `found` flag is set to true in the for loop over implementation requirement's set_parameters. Do we want to reset it before we ran the second loop?

trestle/common/log.py
```diff
@@ -50,7 +50,8 @@ def set_global_logging_levels(level: int = logging.INFO) -> None:
     """
     # This line stops default root loggers setup for a python context from logging extra messages.
     # DO NOT USE THIS COMMAND directly from an SDK. Handle logs levels based on your own application
-    _logger.propagate = False
+    # set propagate to True in order to capture warning messages in tests via caplog
+    _logger.propagate = True
```
enikonovad, line 54: I would avoid changing it here. We can use `pytest capsys` functionality to capture the stdout and stderr output. See `version_test` for an example.

Setting logger to true here may in theory lead to the duplicated messages

enikonovad (approved):
LGTM - I would avoid setting logger.propagate to True as it might pollute the log with duplicated messages. If the only purpose is to test log output then we can use pytest capsys functionality

trestle/common/log.py
fsuits: happy  to use capsys instead so I backed out  the propagate change

trestle/core/control_interface.py
fsuits: The logic is ok because it only falls through to the next loop if found is still false.  There are two places a setparam might hide and only if it isn't in both do you fall through and make a new one.  And even if there is a setparam at the high level - if its value is not the needed value then you need to make a new setparam.

PR #1324, opened 2023-03-16
feat: Rules parameter values validation in SSP

### LINKED ISSUE #1271: Rules parameter values validation in SSP
In a component def. a rule and its parameter values are specified via CSV file. A rule can map to multiple controls, but it has the same value for all controls to which it is mapped. In the CD markdown the rule, parameter and its value are shown in all control Markdown files, to which that rule maps, but the value is non-editable in Markdown as CSV is the source.

But, in an SSP,, the author can specify a different value for a rule parameter than what is there in CD. So, for example if here is a rule "rule1" which has a parameter "param1" with value "1" in CD and is mapped to more than one control say ac-1 and ac-2, then in SSP makrdown both ac-1.md and ac-2.md will have this rule, parameter and its value that is coming from CD. Now, if the SSP author decides to change the value of this rule param to say "2", this needs to be done in both markdown files ac-1.md and ac-2.md. If the user puts different values for same rule parameter in different files then this is invalid.

We need to validate whether the rule parameter settings for each rule param across all controls (to which this rule maps) have the same value. This can be done by having a validation function that performs this check after the SSP is generated in memory and before it is written out as a JSON. 

Work on this has started in branch https://github.com/IBM/compliance-trestle/tree/feat/rules_validator

## Completion Criteria
If the rule param values are not consistent across controls then the ssp-assemble command should return an error with appropriate message and not create the SSP json.

AleJo2995:
## Types of changes

## Summary

Closes #1271 

## Key links:

# Before you merge


fsuits (commented):
The nist-content and nist-source submodules shouldn't be altered by this PR.  But it's possible I had updated them accidentally.  Try running `make nist-default` and then commit/push to see if the change goes away.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class RulesValidator(Validator):
+    """Validator to confirm all rule parameter values are consistent."""
+
+    @staticmethod
+    def val_diff_param_values(existing_param: list, shared_param_values: list) -> list:
+        """
```
fsuits, line 34: typing should be List[str] or something - so we know what it is a list of.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+
+        returns:
+            Difference between values if exist
+        """
+        for param_value in as_list(existing_param):
+            # order params lists and gets differences
+            param_value['values'].sort()
+            shared_param_values.sort()
+            param_vals_set = set(param_value['values'])
+            difference = [x for x in shared_param_values if x not in param_vals_set]
```
fsuits, line 49: If you are just comparing two lists in terms of contents, ignoring order, you can just compare them as ordered lists:  `list1.sort() != list2.sort()`

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+                                        'for the control implementation'
+                                    )
+                            continue
+                        diff = self.val_diff_param_values(existing_params, set_param.values)
+                        if diff:
+                            logger.error(
+                                f'Rule parameter value: {diff} provided for: {set_param.param_id} in '
+                                f'control: {imp_req.control_id} in component: {component.title} is not consistent with '
+                                'other values provided across controls. Invalid model'
+                            )
```
fsuits, line 143: Please check the other validators to see how they indicate VALID or INVALID.  We should keep the messaging structure the same.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+
+        returns:
+            True (valid) if the model's rule parameter values are the same accross controls.
+        """
+        # verify if model type is either an SSP of a Component Definition
+        if not isinstance(model, (SystemSecurityPlan, ComponentDefinition)):
+            return True
+        # iterate by each component defined
+        for component in as_list(model.components):
+            shared_params = {}
```
fsuits, line 69: Only CompDef has components at the top level.   For SSP they are in the system implementation.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+            # iterate by each control implementation within the component
+            for control_imp in as_list(component.control_implementations):
+                _, param_dict, _ = ControlInterface.get_rules_and_params_dict_from_item(control_imp)
+                control_imp_rule_param_ids = [d['name'] for d in param_dict.values()]
```
fsuits, line 73: you also need to check for rule params defined a level above - in the component's properties.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+        for param_value in as_list(existing_param):
+            # order params lists and gets differences
+            param_value['values'].sort()
+            shared_param_values.sort()
+            param_vals_set = set(param_value['values'])
+            difference = [x for x in shared_param_values if x not in param_vals_set]
+            # if there´s a difference in params, log error
+            return difference
+
+    def model_is_valid(self, model: TopLevelOscalModel, quiet: bool) -> bool:
```
enikonovad, line 53: The code in this function is very complex due to the many internal loops and has low reusability. You can split this code into more reusable pieces of code

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+        # iterate by each component defined
+        for component in as_list(model.components):
+            shared_params = {}
+            # iterate by each control implementation within the component
+            for control_imp in as_list(component.control_implementations):
+                _, param_dict, _ = ControlInterface.get_rules_and_params_dict_from_item(control_imp)
+                control_imp_rule_param_ids = [d['name'] for d in param_dict.values()]
+                for set_param in as_list(control_imp.set_parameters):
+                    # if not a rule parameter then ignore
+                    if set_param.param_id not in (control_imp_rule_param_ids):
```
enikonovad, line 76: no need for a brackets around `control_imp_rule_param_ids`. Also you can use `as_filtered_list` to filter elements out.

tests/trestle/core/commands/validate_test.py
```diff
@@ -284,3 +288,71 @@ def test_validate_catalog_params(sample_catalog_rich_controls: Catalog) -> None:
...
+    assem_component, assem_comp_path = model_utils.ModelUtils.load_model_for_class(
+        tmp_trestle_dir, assem_name, comp.ComponentDefinition
+    )
+
+    assert validator.model_is_valid(assem_component, True)
+
+
+def test_rules_validator_unhappy(tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPatch) -> None:
+    """Test validation of rules in catalog."""
+    # FIXME use model with complex rules and param values
```
enikonovad, line 321: Fix this before merging

trestle/core/rules_validator.py
enikonovad: Also quiet is not used. Although it probably shouldn't be used anyways as it would interfere with `-verbose` flag. Use logger levels instead to assign an appropriate level to the message

enikonovad (commented):
Please see below

trestle/core/rules_validator.py
fsuits: The `quiet` is required because it is in the validator ABC signature and it is a command line option specific to validators.  So it's ok here even if not used.

enikonovad: Indeed, forgot we still had this one, thanks for pointing this out

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,145 @@
...
+                        continue
+                    opts_for_parameter = [p['options'] for p in param_dict.values() if p['name'] == set_param.param_id]
+                    shared_params = {
+                        set_param.param_id: [
+                            {
+                                'level': 'ci',
+                                'control_id': '',
+                                'values': set_param.values,
+                                'options': opts_for_parameter
+                            }
```
fsuits, line 86: You are keeping a fairly detailed list of information from all the setparameters - but all that matters is the final value that gets set in each control.  So the only thing you really need to build is a dict of `{param: [{control_id: values}]}`  For each rule param there will be a list of controls that reference that param - and have values assigned by a chain of setparams.  All that matters is the final setparam in that chain - that sets the final value.

AleJo2995:
@fsuits this is no longer valid. Can I just get rid of this or should i keep it for future reference con compdefs validation?

fsuits:
@AleJo2995 yes please close this pr

PR #1330, opened 2023-03-24
chore: Refactor control reader Part 2

enikonovad:
## Types of changes

## Summary
A second part of refactoring of control_reader. This PR touches the process of reading Editable Content from the markdown.
- Refactored a good chunk of parts reading from control_reader
- Removed duplicated code and logic
- Now each node in the tree has a part object in its content (a part is defined by ## in the markdown file)
- Based on the section name a specific format for part_id and part_name will be picked

## Key links:

# Before you merge


trestle/core/markdown/docs_markdown_node.py
```diff
@@ -56,6 +56,14 @@ def __init__(self, key: str, content: DocsSectionContent, starting_line: int):
         super(DocsMarkdownNode, self).__init__(key, content, starting_line)
         self.content: DocsSectionContent = content
 
+    @classmethod
+    def build_tree_from_markdown(cls, lines: List[str], governed_header: Optional[str] = None):
+        """Construct a tree out of the given markdown."""
+        ob = cls.__new__(cls)
+        start_level = ob._get_max_header_lvl(lines)
```
fsuits, line 63: I see this `__new__` has been in the code for some time - but it is the only time it is used.  Just want to confirm there is a reason this classmethod is special and needs it.

trestle/core/markdown/control_markdown_node.py
```diff
@@ -55,10 +86,8 @@ class ControlSectionContent(BaseSectionContent):
     def __init__(self):
         """Initialize section content."""
         super(ControlSectionContent, self).__init__()
-        self.statement_part = None
-        self.objective_part = None
-        self.guidance_part = None
-        self.other_parts = []
+        self.a_part = None
+        self.by_id_name = ''
```
fsuits, line 90: Calling it `a_part` is odd to me.  Partly because a statement could have `Part a.` in it.  Is there a better name to describe what it is and what it contains?  Some doc text would help also.

trestle/core/markdown/base_markdown_node.py
```diff
@@ -148,14 +148,8 @@ def delete_nodes_text(self, keys: List[str], strict_matching: bool = True) -> Li
...
-    def _build_tree(
-        self,
-        lines: List[str],
-        root_key: str,
-        starting_line: int,
-        level: int,
-        governed_header: Optional[str] = None
-    ) -> Tuple[BaseMarkdownNode, int]:
+    def _build_tree(self, lines: List[str], root_key: str, starting_line: int,
+                    level: int) -> Tuple[BaseMarkdownNode, int]:
```
fsuits, line 152: I think this function is dead and is not being called.  `level` isn't used and the function has no code coverage.  But it still may be needed in terms of the overall API...

trestle/core/markdown/control_markdown_node.py
```diff
@@ -17,19 +17,46 @@
...
+    """
+
+    control_statement = 1
+    control_objective = 2
+    control_guidance = 3
+    editable_by_id_part = 4
+    generic_control_part = 5
+    generic_subpart = 6
+    editable_content = 7
+    undefined = 8
```
fsuits, line 57: Please use caps for enums.  I think it's fairly standard convention and it is the convention in trestle - except in the auto-generated OSCAL python classes.  And it would be good to add doc strings on Part vs. part - e.g. is `## part foo` the same as `## Part foo`

trestle/common/const.py
```diff
@@ -228,7 +228,7 @@
 # need to strip white space from end of value
 PROPERTY_REGEX = r'(?:###\s+Property\s+)([^\s]*)\s*:\s+(.*)'
 
-PART_REGEX = r'(?:##\s+Part\s+)(.*)'
+PART_REGEX = r'(?:##\s+part\s+)(.*)'
```
fsuits, line 232: I think it's worth changing this to have `(?i)` in front to indicate directly that case is ignored when matching the string.  Then you can remove all the `lower()` references in the code that do matching.  Otherwise if a `lower()` is omitted it could be a bug.

trestle/core/markdown/control_markdown_node.py
```diff
@@ -145,70 +165,51 @@ def _add_child_prose_if_need(part: common.Part, text: str, should_add: bool):
...
-
-                    part_id = ControlInterface.strip_to_make_ncname(
-                        tree_context.control_id + '_' + control_md_heading_label
-                    )
-                    a_part = common.Part(id=part_id, name=control_md_heading_label_ncname, prose='')
-                    content.other_parts.append(a_part)
-
-                i = self._process_part_line(i, line, lines, content.other_parts[-1], read_parts=False)
+            if section_heading_type not in [ControlSectionType.undefined, ControlSectionType.editable_content]:
+                # Read part
```
fsuits, line 204: Should we have debug message or warning for undefined type?  It may represent content the user meant as an important part and it is being dropped - so I think a warning.

trestle/core/markdown/control_markdown_node.py
```diff
@@ -239,6 +236,126 @@ def get_control_guidance(self) -> Optional[ControlMarkdownNode]:
...
+        editable_node = self.get_node_for_key(f'# {const.EDITABLE_CONTENT}')
+        parts = self._get_subparts(editable_node, exclude_parts_id)
+        return parts
+
+    def _get_subparts(self,
+                      control_node: ControlMarkdownNode,
+                      exclude_ids: List[str] = None) -> Optional[List[common.Part]]:
+        """Get subparts of the control part node if exists."""
+        if not control_node:
+            raise TrestleError('No control node was provided to extract subparts.')
```
fsuits, line 269: Is there a way to provide more context for this error so the user knows where the problem is?  Perhaps the check needs to be before this call.

fsuits (approved):
LGTM after addressing comments - many of which are style and naming.  Before merge I think we should go over the terms and make sure they are used correctly: part, section and guidance.  So please hold until after discussion.

trestle/core/markdown/docs_markdown_node.py
enikonovad: This is a fairly special use case for this class. This is a classmethod factory method to create the instance of MarkdownNode. The reason for this design choice was that each MarkdownNode instance operates on itself and builds the tree recursively. It needs to perform some subroutines (i.e. compute the max heading level, key, etc) and contain some information in each node (defined in __init__). Because we don't want to call `__init__` before we compute those this approach was used

trestle/core/markdown/control_markdown_node.py
enikonovad: `a_part` could represent any part from the markdown. A part is something that starts with two or more hashes. Let me think if there is any other generic enough name for this

trestle/core/markdown/base_markdown_node.py
enikonovad: Good catch, I meant to replace this with `pass` as child classes should implement it

trestle/common/const.py
enikonovad: Good suggestion, didn't think of modifying regex. Will adjust

PR #1337, opened 2023-03-31
feat: validate SSP rule parameter values

### LINKED ISSUE #1271: Rules parameter values validation in SSP

AleJo2995:
## Types of changes

## Summary

Closes #1271 

## Key links:

# Before you merge


tests/trestle/core/commands/validate_test.py
```diff
@@ -284,3 +291,113 @@ def test_validate_catalog_params(sample_catalog_rich_controls: Catalog) -> None:
...
+
+    orig_ssp, _ = ModelUtils.load_model_for_class(tmp_trestle_dir, ssp_name, ossp.SystemSecurityPlan)
+
+    by_components = orig_ssp.control_implementation.implemented_requirements[1].by_components
+    set_parameter = orig_ssp.control_implementation.implemented_requirements[0].by_components[0].set_parameters[1]
+    props = orig_ssp.control_implementation.implemented_requirements[0].by_components[0].props
+    by_components[0].props = props
+    by_components[0].set_parameters = []
+    by_components[0].set_parameters.append(set_parameter)
+    by_components[0].set_parameters[0].values.append('shared_param_1_ab_opt_2')
```
fsuits, line 332: I believe this is passing a reference rather than a copy.  As a result you are changing the values in both places and the result will be valid - which is not what you want.  I think making a copy of the set_param that you are adding will fix it - since the test is failing because the model is showing as valid.

fsuits: A short comment describing what you are doing would be good.  It should be easy to make it clear that you are copying a setparam fron one bycomp so you can then set its value to something different.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,110 @@
...
+from trestle.common.list_utils import as_list, deep_get, deep_set, get_item_from_list
+from trestle.core.catalog.catalog_interface import CatalogInterface
+from trestle.core.profile_resolver import ProfileResolver
+from trestle.core.validator import Validator
+from trestle.oscal.ssp import SystemSecurityPlan
+
+logger = logging.getLogger(__name__)
+
+shared_params = {}
+
```
fsuits, line 32: Please avoid using globals unless there is a very good reason you want to share it throughout the code - which is very rare.  Also - I would call this `rule_param_dict` as it is only called shared params in the particular data set.  The intent here is to gather all rule_params and associated values - and then check for any that have multiple different values across controls.  It is also a dict rather than a simple list, so I would make that clear.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,110 @@
...
+
+class RulesValidator(Validator):
+    """Validator to confirm all rule parameter values are consistent."""
+
+    def add_shared_param(
+        self, item: TypeWithProps, control_id: str, cat_int: CatalogInterface, comp_name: str = ''
+    ) -> None:
+        """
```
fsuits, line 40: This actually needs an item with SetParameters - so I would create a new common type called TypeWithSetParameters in `common_types.py`

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,110 @@
...
+        Add a rule shared parameter to the rule shared parameters list.
+
+        args:
+            item: Generic item to iterate over parameters.
+            control_id: Current control id.
+            cat_int: Instance of catalog interface with controls catalog loaded.
+            comp_name: Component name to save.
+        """
+        for param in as_list(item.set_parameters if not isinstance(item, cat.Control) else item.params):
+            parameter_id = param.param_id if not isinstance(item, cat.Control) else param.id
```
fsuits, line 50: no need to check controls for params.  use the catalog interface to see if a given param_id is in the catalog.  If it is not, then assume it is a rule_param.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,110 @@
...
+            return True
+
+        components = as_list(model.system_implementation.components)
+        component_titles = [x.title for x in components]
+        profile_resolver = ProfileResolver()
+        catalog = profile_resolver.get_resolved_profile_catalog(trestle_root, model.import_profile.href)
+        catalog_interface = CatalogInterface(catalog)
+
+        for control in catalog_interface.get_all_controls_from_catalog(True):
+            self.add_shared_param(control, control.id, catalog_interface)
```
fsuits, line 78: no need for this.  when you add the param, check to see if it is known in the catalog interface by param_id.  If it is known then ignore it and do not add it to the dict.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,110 @@
...
+                    component = get_item_from_list(components, by_component.component_uuid, lambda x: x.uuid)
+                    self.add_shared_param(by_comp, imp_req.control_id, catalog_interface, component.title)
+        if not shared_params:
+            return True
+        # compare all values in shared paramerets by component basis
+        for shared_param, _ in as_list(shared_params.items()):
+            for comp_title in component_titles:
+                # retrieve all control rule param values attached to a component
+                controls_list = deep_get(shared_params, [shared_param, comp_title], None)
+                if not controls_list:
```
fsuits, line 98: This overall logic can be simplified.  The dict is a dict of `{paramid: {comp_id: {control_id: values}}}`  so just grab all its values of the form `{comp_id: {control_id: values}}` and the values for each control id should be the same.

tests/trestle/core/commands/validate_test.py
AleJo2995: Got it, addressed

trestle/core/rules_validator.py
AleJo2995: changed.

trestle/core/rules_validator.py
AleJo2995: Done. There was actually already a TypeWithSetParameters created so I used it. thanks

trestle/core/rules_validator.py
AleJo2995: got it.

trestle/core/rules_validator.py
AleJo2995: done resolved

trestle/core/rules_validator.py
AleJo2995: done. resolved

fsuits (commented):
Looks much better but it still needs to hook into ssp-assemble so that just prior to writing out, after assembly, the ssp is checked for validity.  If it is not valid it should give warning and not write it out - with indication of what the conflict was (what param and what control - etc).  There should also be a test for this situation.

AleJo2995:
So @fsuits , I´ve tried to implement the validation before writing out the SSP, but what I don´t know is what should I do for all other test cases (e.g `test_ssp_get_control_response`) that assume that the SSP is assembled. Does this mean that I need to modify all test cases that use ssp_assemble and adjust them so they wont fail and then create a test for this particular scenario?

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,119 @@
...
+            self.iter_and_add(imp_req.by_components, components, catalog_interface, imp_req)
+            # includes rule param values in each by_component present in statements
+            for statement in as_list(imp_req.statements):
+                self.iter_and_add(statement.by_components, components, catalog_interface, imp_req, statement)
+        if not self.rule_param_values_dict:
+            return True
+        # compare all values in shared paramerets by component basis
+        for shared_param, values_dict in as_list(self.rule_param_values_dict.items()):
+            for comp_name, value_dict in values_dict.items():
+                expected_value = next(iter(value_dict.values()))
```
enikonovad, line 111: Not sure if initialising iterator here is needed?

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,119 @@
...
+from trestle.core.validator import Validator
+from trestle.oscal.ssp import ByComponent, ImplementedRequirement, Statement, SystemComponent, SystemSecurityPlan
+
+logger = logging.getLogger(__name__)
+
+
+class RulesValidator(Validator):
+    """Validator to confirm all rule parameter values are consistent."""
+
+    rule_param_values_dict = {}
```
enikonovad, line 35: Now I would be very careful with the class attributes. Class attributes are shared across all instances of the RulesValidator. What it means is that this dictionary will be never reset between validating different ssps. Is this intended behaviour?

fsuits:
@AleJo2995 no - all you need to do is test that it works for one invalid ssp assembly.  All the other tests should inherently test the good path.  Your recent tests test the internal function itself - and you just need an additional test that catches one bad ssp before it is written out.

enikonovad (commented):
Some minor typos and comment, please check that class attribute is really needed here

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,119 @@
...
+            control_id: Current control id.
+            cat_int: Instance of catalog interface with controls catalog loaded.
+            comp_name: Component name to save.
+        """
+        for set_param in as_list(item.set_parameters):
+            control = cat_int.get_control_by_param_id(set_param.param_id)
+            if not control:
+                deep_set(self.rule_param_values_dict, [set_param.param_id, comp_name, control_id], set_param.values)
+
+    def model_is_valid(self, model: TopLevelOscalModel, quiet: bool, trestle_root: pathlib.Path = '') -> bool:
```
enikonovad, line 80: Sorry I have just realised there is one major issue with this PR.

`trestle validate` does not pass `trestle_root` to this validator. Please see line 71 in `validator.py`.

Here you initialise it to the empty string. This will fail when user runs `trestle validate` and no guidance would be provided to the user on how to fix this issue.

You can either take a current repository as the root or raise an exception to notify user that root is required. In any case even if the user provides the `--trestle_root` it will not be passed.

I would recommend creating a test that would run `trestle validate` on the ssp repository using `execute_command_and_assert` to catch this. 

Also please give it a test on `ocp-oscal-ssp`

enikonovad (changes requested):

trestle/core/rules_validator.py
fsuits: Thanks @enikonovad this is something special to validating the ssp that is a break from the normal validation calls.  @AleJo2995 sorry I didn't anticipate this issue, but it shouldn't be too hard to fix.  Normally a model in memory can be validated without trestle root because it doesn't need to load anything from the trestle workspace.  But the ssp validation is special since it needs to load the profile.  I think the simplest fix is to change the signature of `model_is_valid` in the base class so it passes in an Optional trestle root as `pathlib.Path` or None.  None of the other validators need trestle_root but the ssp one does.  Also not that pathlib.Path is a special class and is not a string - so if not present it should default to None.

trestle/core/rules_validator.py
fsuits: yes please don't place things at the top as attributes of the class.  Simplest is simply to define it in `model_is_valid` and pass as argument to the other methods.  You may also be able to add an `__init__` to the class and add it as a private member that the other functions refer to.

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,119 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class RulesValidator(Validator):
+    """Validator to confirm all rule parameter values are consistent."""
+
+    rule_param_values_dict = {}
+
+    def iter_and_add(
```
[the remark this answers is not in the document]
AleJo2995: Done

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,119 @@
...
+        self,
+        by_components: List[ByComponent],
+        components: List[SystemComponent],
+        cat_int: CatalogInterface,
+        imp_requirement: ImplementedRequirement,
+        statement: Statement = None
```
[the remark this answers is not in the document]
AleJo2995: done

trestle/core/rules_validator.py
```diff
@@ -0,0 +1,119 @@
...
+        """
+        Test if the model is valid.
+
+        args:
+            model: A top level OSCAL model.
+            quiet: Don't report msgs unless invalid.
+            trestle_root: Trestle root path.
+
+        returns:
+            True (valid) if the model's rule parameter values are the same accross controls.
```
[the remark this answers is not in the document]
AleJo2995: done

trestle/core/rules_validator.py
AleJo2995: done

AleJo2995:
@fsuits for validating an ssp has no given profile I´ve created `test_validate_ssp_with_no_profile` . Please check. Actually I wanted to add as well a check for `trestle validate` command and `execute_command_and_assert` to test validator executes what is expected, but it seems that for current ssp there seem to be duplicate ids so the validator is failing even before reaching the point to validate rules.

This is the code I´m trying to execute:

`validate_command = f'trestle validate -t system-security-plan -n {ssp_name}'
    test_utils.execute_command_and_assert(validate_command, 0, monkeypatch)`

fsuits:
@AleJo2995 Sorry for the test fail.  @enikonovad also noticed this problem today and created an issue.  This is an old bug that the validator is complaining about duplicates that are in fact ok.  The check for duplicate param_ids should not happen in comp defs and ssps.  I should have a fix today.

tests/trestle/core/commands/validate_test.py
```diff
@@ -280,7 +288,116 @@ def test_validate_catalog_params(sample_catalog_rich_controls: Catalog) -> None:
...
+
+    # first create the markdown
+    ssp_gen = SSPGenerate()
+    assert ssp_gen._run(gen_args) == 0
+
+    # first ssp assembly
+    ssp_assemble = f'trestle author ssp-assemble -m {ssp_name} -o {ssp_name} -cd {args_compdefs}'
+    test_utils.execute_command_and_assert(ssp_assemble, 0, monkeypatch)
+
+    new_ssp, _ = ModelUtils.load_model_for_class(tmp_trestle_dir, ssp_name, ossp.SystemSecurityPlan)
```
enikonovad, line 315: This code block repeats 3 times in the tests, please create a helper function to not duplicate the code.

trestle/core/commands/author/ssp.py
```diff
@@ -554,7 +556,16 @@ def _run(self, args: argparse.Namespace) -> int:
             if args.regenerate:
                 ssp, _, _ = ModelUtils.regenerate_uuids(ssp)
             ModelUtils.update_last_modified(ssp)
-
+            # validates if ssp doesn´t have a profile a provides warning
+            if not ssp.import_profile.href:
+                logger.warning('No profile given for current SSP. Model is invalid')
```
enikonovad, line 561: Please use `logger.error` for this and next warning. Since program quits after this log it should be an error

trestle/core/commands/common/return_codes.py
```diff
@@ -41,3 +42,4 @@ class CmdReturnCodes(enum.Enum):
     IO_ERROR = 6
     AUTH_ERROR = 7
     UNKNOWN_ERROR = 8
+    INVALID_MODEL = 9
```
enikonovad, line 45: Please use `OSCAL_VALIDATION_ERROR ` as it was created for this purpose

tests/trestle/core/commands/validate_test.py
AleJo2995: Done

trestle/core/commands/author/ssp.py
AleJo2995: Done

trestle/core/commands/common/return_codes.py
AleJo2995: done

tests/trestle/core/commands/author/jinja_cmd_test.py
```diff
@@ -25,14 +25,14 @@
 from trestle.core.markdown.docs_markdown_node import DocsMarkdownNode
 
 
-def setup_ssp(testdata_dir: pathlib.Path, tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPatch):
+def setup_ssp(testdata_dir: pathlib.Path, tmp_trestle_dir: pathlib.Path, monkeypatchx: MonkeyPatch):
     """Prepare repository for docs generation."""
```
fsuits, line 29: please change this back to monkeypatch

tests/trestle/core/commands/validate_test.py
```diff
@@ -296,7 +303,107 @@ def test_validate_catalog_params(sample_catalog_rich_controls: Catalog) -> None:
...
+    test_utils.execute_command_and_assert(validate_command, 4, monkeypatch)
+
+    new_ssp.import_profile.href = original_href
+    ModelUtils.save_top_level_model(new_ssp, tmp_trestle_dir, ssp_name, FileContentType.JSON)
+
+
+def test_rule_param_values_validator_unhappy(tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPatch) -> None:
+    """Test validation of rules in SSP."""
+    args = argparse.Namespace(mode=const.VAL_MODE_RULES)
+    validator: Validator = validator_factory.get(args)
```
fsuits, line 360: pls move this test up to be next to the happy one

tests/trestle/core/commands/validate_test.py
```diff
@@ -296,7 +303,107 @@ def test_validate_catalog_params(sample_catalog_rich_controls: Catalog) -> None:
...
+    # adds a new set parameter to set parameters array for current by component in statement
+    by_components[1].set_parameters.append(new_set_parameter)
+    imp_reqs = new_ssp.control_implementation.implemented_requirements
+    imp_req_at = next((i_req for i_req in imp_reqs if i_req.control_id == 'at-1'), None)
+    by_components = imp_req_at.statements[0].by_components
+    by_components[0].set_parameters = []
+    new_set_parameter_one = gens.generate_sample_model(ossp.SetParameter)
+    new_set_parameter_one.values = ['shared_param_1_ab_opt_1']
+    new_set_parameter_one.param_id = 'shared_param_1'
+    by_components[0].set_parameters.append(new_set_parameter_one)
```
fsuits, line 404: There are multiple reuses of variable names here: by_components, imp_reqs, etc.  I think they are ok but I would keep the new_set_parameters separate.  Pls call the first one new_set_parameter_a and the other new_set_parameter_b

enikonovad (approved):
LGTM - please address minor comments before merging and double check with @fsuits if he also approves

trestle/core/rule_parameters_validator.py
```diff
@@ -0,0 +1,125 @@
...
+        args:
+            by_components: A list of by components.
+            components: Components defined at System implementation level.
+            cat_int: Instance of catalog interface with controls catalog loaded.
+            imp_requirement: Current implemented requirement.
+            statement: Statement if iterating by statements.
+        """
+        for by_component in as_list(by_components):
+            component = get_item_from_list(components, by_component.component_uuid, lambda x: x.uuid)
+            # first adds implmented requirement set params
```
fsuits, line 59: raise exception if component is None

trestle/core/rule_parameters_validator.py
```diff
@@ -0,0 +1,125 @@
...
+            if not statement:
+                self._add_shared_param(imp_requirement, imp_requirement.control_id, cat_int, component.title)
+            # then  adds by components set params per requirement
+            self._add_shared_param(by_component, imp_requirement.control_id, cat_int, component.title)
+
+    def _add_shared_param(
+        self, item: TypeWithSetParams, control_id: str, cat_int: CatalogInterface, comp_name: str = ''
```
fsuits, line 66: call this _add_rule_params because it is adding all rule params and they may not be shared

trestle/core/rule_parameters_validator.py
```diff
@@ -0,0 +1,125 @@
...
+
+        args:
+            item: Generic item to iterate over parameters.
+            control_id: Current control id.
+            cat_int: Instance of catalog interface with controls catalog loaded.
+            comp_name: Component name to save.
+        """
+        for set_param in as_list(item.set_parameters):
+            control = cat_int.get_control_by_param_id(set_param.param_id)
+            if not control:
```
fsuits, line 79: add comment here explaining that if the param is not associated with control we assume it is a rule param

trestle/core/rule_parameters_validator.py
```diff
@@ -0,0 +1,125 @@
...
+        imp_requirement: ImplementedRequirement,
+        statement: Optional[Statement] = None
+    ) -> None:
+        """
+        Iterate all by components in an object and add the rule shared parameter values to list.
+
+        args:
+            by_components: A list of by components.
+            components: Components defined at System implementation level.
+            cat_int: Instance of catalog interface with controls catalog loaded.
```
fsuits, line 53: You don't really need the components list because you can build the dict just using the comp_uuids.  So this can be simplified.  The comp title is useful for a debug or warning msg but not essential

trestle/core/rule_parameters_validator.py
```diff
@@ -0,0 +1,125 @@
...
+
+        components = as_list(model.system_implementation.components)
+        if not model.import_profile.href:
+            logger.info(f'INVALID: Model {model.metadata.title} has no referenced profile')
+            return False
+        profile_catalog = ProfileResolver().get_resolved_profile_catalog(trestle_root, model.import_profile.href)
+        catalog_interface = CatalogInterface(profile_catalog)
+        # iterate by each implemented requirement defined
+        for imp_req in as_list(model.control_implementation.implemented_requirements):
+            # iterate by each by_component in each implemented requirement
```
fsuits, line 108: you don't need as_list here because imp_reqs are required

trestle/core/rule_parameters_validator.py
```diff
@@ -0,0 +1,125 @@
...
+            self._iter_and_add(imp_req.by_components, components, catalog_interface, imp_req)
+            # includes rule param values in each by_component present in statements
+            for statement in as_list(imp_req.statements):
+                self._iter_and_add(statement.by_components, components, catalog_interface, imp_req, statement)
```
fsuits, line 112: You are calling iter_and_add twice but inherently for the same imp_req.  Instead you could put this logic together in a separate function and just call it _add_imp_req_setparams_to_dict or something.  I don't think there is a need for the optional statement argument.

tests/trestle/core/commands/author/jinja_cmd_test.py
AleJo2995: done

tests/trestle/core/commands/validate_test.py
AleJo2995: done

tests/trestle/core/commands/validate_test.py
AleJo2995: done

trestle/core/commands/author/ssp.py
```diff
@@ -554,7 +556,12 @@ def _run(self, args: argparse.Namespace) -> int:
             if args.regenerate:
                 ssp, _, _ = ModelUtils.regenerate_uuids(ssp)
             ModelUtils.update_last_modified(ssp)
-
+            # validate model rules before saving
+            args_validate = argparse.Namespace(mode=const.VAL_MODE_RULES)
+            validator: Validator = validator_factory.get(args_validate)
+            if not validator.model_is_valid(ssp, True, trestle_root):
+                logger.warning('Validation of file to be imported did not pass.  Import failed.')
+                return CmdReturnCodes.COMMAND_ERROR.value
```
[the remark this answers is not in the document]
AleJo2995: done

trestle/core/rule_parameters_validator.py
AleJo2995: done

trestle/core/rule_parameters_validator.py
AleJo2995: done

trestle/core/rule_parameters_validator.py
AleJo2995: done

trestle/core/rule_parameters_validator.py
AleJo2995: done

trestle/core/rule_parameters_validator.py
AleJo2995: done

trestle/core/rule_parameters_validator.py
AleJo2995: work in progress...

AleJo2995: Done and tested.

fsuits (approved):
LGTM

Issue #1755, opened 2024-11-19
Enable running unit tests in PyCharm


d10n:
## Issue description / feature objectives
* PyCharm has quirks related to the current working directory when running and debugging tests.
* Tests use `pathlib.Path('tests/data')` frequently. Where does that resolve the path from? Relative to the current working directory.
* Where is the `tests/data` directory? Relative to the tests.
* Therefore, PyCharm does not have integrated test running and debugging.

## Caveats / Assumptions
-
## Completion Criteria

* Unit tests should pass in the PyCharm test runner, and in all the execution environments they originally passed in too.

PR #1756, opened 2024-11-19
fix: unit test execution in PyCharm (#1755)

d10n:
* Where does pathlib.Path('tests/data') resolve the path from? It's
  relative to the current working directory.
* Where is the tests/data directory? Relative to the tests.
* PyCharm has quirks related to the current working directory when
  running and debugging tests.
* By using __file__ to resolve the test data directory, we remove the
  non-guaranteed assumption that the current working directory is always
  at the root of the repository, and we replace that assumption with a
  value that is always correct regardless of current working directory.
* As a result, the unit tests can now be debugged in PyCharm, unit tests
  can be executed from any directory, and test execution continues to
  work in all the places it originally worked.

Some of the handling is slightly messy but that can be cleaned up by
making the task config files themselves not dependent on the CWD

## Types of changes

## Key links:

# Before you merge

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -17,6 +17,7 @@
 
 import argparse
 import difflib
+import functools
 import logging
 import os
 import pathlib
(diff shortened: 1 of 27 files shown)


d10n:
PyCharm has docs on creating the unit test configuration, but I'll add a short blurb about it to the docs here too

degenaro:
What I'm after is a way to reproduce the problem and show that it has been fixed. I installed PyCharm (community edition). I downloaded the trestle repo (develop branch). I can Run and Debug for a test script (cis_xlsx_to_oscal_catalog.py), though the for Debug case when I place breakpoints they do not seem to cause a stop - the console message is `Process finished with exit code 0`. Sorry to be a PyCharm newbee here...

d10n:
Thanks for the bump - I'll record a short screen recording and attach it here

d10n:
I just realized that I rebased against an old develop branch (thanks @jpower432)

tests/test_utils.py
```diff
@@ -53,31 +54,54 @@
...
+            try:
+                f(*args, **kwargs)
+            finally:
+                os.chdir(original_cwd)
+
+        return wrapper
+
+    return decorator
+
+
```
degenaro, line 104: `Not safe for concurrent tests which may change directory.` This is somewhat concerning. Running `make test` causes a set of parallel chunks of tests to be run. It is good that all the tests passed for this PR, but was that luck or not? Thx.

d10n: I believe that this is practically a no-op in the case of `make test` because the decorator `cd`s to the root directory and `make test` only works in the root directory anyway, so the decorator starts at the test root, then cds to the test root, then cds back to the test root after the test runs.
It's good that this unsafe workaround triggers raises a mental flag because it is only needed on tests that have an implicit dependency on paths relative to the test root directory

degenaro: OK. I suggest adding similar comments in the code itself.

jpower432 (commented):
This LGTM, but it looks like the last change is failing the formatting check.

jpower432:
@d10n I think this PR is ready to go. Could you please apply the formatting fixes to make sure the check is passing? Running `make code-format` locally should do the trick.

jpower432 (approved):
LGTM

PR #2041, opened 2026-01-16
feat: add a codegen test

butler54:
Signed-off-by: Chris Butler <chris.butler@redhat.com>


butler54:
That we have not enforced yet and is GH configuration. At this state the answer is - no it's not a requirement.

PR #2095, opened 2026-02-26
fix(build): fix mkdocs htmlproofer link validation flakiness

### LINKED ISSUE #2032: mkdocs linting is becoming flaky in CICD testing and may affect releases
## Describe the bug

the `lint` job run as part of the CICD process in github actions is intermittently failing. The errors appear to be due to 503's on links in mkdocs which are being explored

## To Reproduce

setup a PR and repeatedly re-run the tasks until a failure occurs (at current rate every 2-3 times)

## Expected behavior

Flakiness should not occur at this rate. Explore reasons why / add retries. 

## Environment

- OS: \[e.g. iOS\]
- Python version:
- Installed packages:

khushiiagrawal:
## Types of changes

## Summary
Resolves #2032 

The docs validate step in the CI pipeline is intermittently failing due to 503 Service Unavailable responses from valid external links.

This PR fixes the flakiness by configuring the mkdocs-htmlproofer-plugin to ignore 503 (Service Unavailable) and 504 (Gateway Timeout) HTTP status codes natively. Since these are temporary server-side issues rather than broken links in our documentation, explicitly excluding them from the failure checks stabilizes the CI without losing the benefits of validating 404s.


degenaro (commented):
@khushiiagrawal Thanks for this PR. It seems that 503 and 504 will now be ignored. Is there a warning about such URLs so that the PR reviewer can judge whether this of concern or not? Are there any retrys before the ignore/warning happens?

butler54:
> This will cause 503/504 to always be ignored in the github pipeline.
> 
> I don't suppose there's a way when running the Makefile locally on my laptop to not ignore 503/504, unless the yml file is hacked?
> 
> Do we know if the URLs currently in the ignore list still fail and if so is it because of 503/504?
> 
> Funny, I just enabled the pipelines to run and there was an htmlproofer failure with 502.

I think we can do environmental variable overrides in the mkdocs.yml. You can see an example of those changes Here;
  https://github.com/aucloud/aucloud.github.io/blob/0c9d2012825e87318e58294c299bc06d5393acb6/mkdocs.yml#L62-L67
```yaml
  - htmlproofer:
      enabled: !ENV [ENABLED_HTMLPROOFER, False]
      validate_rendered_template: False
      validate_external_urls: False
      raise_error_after_finish: False
      raise_error: False
  ```
  
  we'd need to chain that into the makefile.

degenaro:
@khushiiagrawal In the end the website should have not have broken links. There may be times when these sites are not available for a variety of reasons and not much can be done about that.

We should not err on the side accepting broken links (external or internal) for the sake of expediency. The whole idea of checking is that we have not introduced a new link that is broken (or happen to come across an existing link that has since broken).

degenaro:
Fixed by #2262 (for now).

PR #2111, opened 2026-03-03
fix: correct metadata/version check in version command

rishi-jat:
## Issue

While reviewing the code, I noticed the boolean condition in version.py:

    if not (oscal_object.metadata or oscal_object.metadata.version)

This can raise an AttributeError when metadata is None and does not correctly validate missing version values.

## Changes

- Updated the condition to explicitly check metadata and version separately.
- Added a unit test to ensure the command raises TrestleError when metadata is None.

Prevents a potential crash and improves validation logic.


butler54 (changes requested):
I think we need to do some more refactoring here:

TL;DR - i'm not convinced you can hit the code properly anyway. Why?
The top level OSCAL models have metadata as mandatory
e.g. https://github.com/oscal-compass/compliance-trestle/blob/87fe40ba2d417601b525b05f64bf7679fa154a89/trestle/oscal/poam.py#L111

so Metadata, with pydandic enforcement, cannot be none.

Issue #1959, opened 2025-10-21
Use NIST 800-53 rev 5


degenaro:
## Issue description / feature objectives

NIST 800-53 Revision 4 was officially withdrawn on September 23, 2021. Survey the use of Rev 4 in the core, demo, and associated OSCAL Compass repos. Switch to Rev 5 when appropriate (most if not all cases).

## Caveats / Assumptions

Create individual issues to divide and conquer.

## Completion Criteria

Set of issues (e.g. one or more for trestle, one or more for demos...) created such that use of NIST 800-53 Revision 4 will be minimized or eliminated.

degenaro:
@Prarthana-1910 I believe sub issues #1977, #1978 have been closed by PRs. If there are other places that rev4 appears, please open a sub issue. Else, this epic issue can be closed as well. Thx!

degenaro:
Fixed by #2116

PR #2116, opened 2026-03-06
fix: switch from rev4 to rev5 in test files

LAKSHJAIN14:
## Types of changes

## Summary

As discussed with @degenaro , this PR fixes a part of #1959 

switched from rev4 to rev5 and removed some constants in the `/tests/test_utils.py` as they were no longer being used anywhere and their equivalent rev5 constants were already there so just wanted to know if there's no problem in that ?


degenaro:
@LAKSHJAIN14 Yes, there is some work involved with converting from rev4 to rev5. Do you have an example of one that you are having trouble with?

degenaro:
AI should be able to help with: `change the test data to be mapped against rev5 controls`. Thx!

LAKSHJAIN14:
Yeah Ig I would do that, I thought you might have an existing test data that was mapped to rev5

LAKSHJAIN14:
@degenaro 
Have made the changes in the test data to be mapped against rev 5 controls USING AI and also changed the source URL to point at the rev5 catalog

Also the `bad_component_dup_uuid.yaml` file is not being used in any test or anything so its a dead data like I mentioned in #2142 so do tell me if we want that file to be removed.

degenaro:
@LAKSHJAIN14 Looks mostly good. Lower case seems correct for the labels, but why are the parts dropped such as .a, .b?

image

degenaro:
> Also the `bad_component_dup_uuid.yaml` file is not being used in any test or anything so its a dead data like I mentioned in #2142 so do tell me if we want that file to be removed.

Yes, if not used it should be deleted. Thx!

degenaro:
@vikas-agarwal76 Could you comment of the ac-8 vs. ac-8.a? (Are part controls?)

PR #2118, opened 2026-03-06
fix: correct error path in assemble command

rishi-jat:
### Issue

The assemble command checks root_model_filepath but the error message
references root_model_dir, which can be misleading.

### Changes

Update the error message to reference root_model_filepath so the correct
path is shown when the file is missing.


degenaro:
Claude review:

The stated fix (last hunk) is good. Simple, correct, matches the title/description exactly — it now reports the actual file path instead of the directory.

There's a second, undocumented change (the new try/except around get_contextual_file_type) that the PR description doesn't mention. A few things worth flagging in review:

Scope creep vs. description — the PR body only talks about fixing "the error message," but this hunk adds new exception-handling behavior for a different failure mode (empty/missing model directory). Worth calling out explicitly in the PR description so reviewers know two things changed.
Swallows a more specific error — get_contextual_file_type can raise TrestleError for two different reasons: "Trestle workspace not found at path {path}" or "No files found in the project." The new except TrestleError catches both and replaces them with a generic "No top level model file at ..." message, losing the more specific "not a valid trestle workspace" diagnostic in that first case.
No exception chaining — raise TrestleError(...) inside the except block drops the original traceback. Using raise TrestleError(...) from e would preserve it for debugging.
Hardcoded .json extension — the synthesized path in the except block always assumes .json (f'{model_alias}.json'), even though projects can use yaml/xml. Minor since it's only used for a human-readable error message, but slightly inconsistent.
Test coverage is good — test_assemble_missing_root_model_file_raises exercises the new branch, though it only asserts a substring match ('No top level model file at'), not the full path, so it wouldn't catch a regression in path construction.

Net: the core fix is solid and low-risk; I'd ask the author to either narrow the except to the specific "no files found" case (or chain the exception) and update the PR description to mention this second behavioral change.

degenaro (approved):
LGTM

PR #2120, opened 2026-03-06
fix: preserve falsy values in list_utils helpers

rishi-jat:
### Issue

Some helpers in list_utils.py use truthiness checks when setting dictionary
values. This causes valid falsy values such as 0, False, or "" to be removed.

### Changes

Replace truthiness checks with explicit None checks so falsy values are
preserved while None still removes the key.


degenaro:
image

PR #2133, opened 2026-03-10
feat: add RFC 8785 (JCS) canonicalization support for OSCAL JSON docu…

### LINKED ISSUE #2013: Support JSON Canonicalization Scheme to support cryptographic signing scenarios
In investigating the opportunity for using signing in the oscal release process one of the issues is verifiability. For verifiability binary exact outputs need to be repeatable, ideally across systems and implementations.

e.g. given an oscal json object there are potentially different ways this can be represented even as json.**

Json canonicalization scheme [RFC 8785](https://www.rfc-editor.org/rfc/rfc8785) is a method to get a repeatable output for a given functional.

1. Introduce RFC8785 formatted json (aka JCS) as a supported output type.
2. Introduce flags throughout  CLI commands to support this.
3. Introduce a command to allow object canonicalisation.

** note yaml does not have a approved normalisation standard [see here](https://github.com/yaml/yaml-spec/issues/289) so we have excluded for now.

## Caveats / Assumptions
- YAML is out of scope
- XML is out of scope
- The `jcs` library is out of scope
- This is purely canonicalisation. It will *not* examine other aspects of the build process which are required to get [reproducible builds](https://reproducible-builds.org/), however, it's addresses a core concern.

Acuspeedster:
# feat: Add RFC 8785 (JCS) canonicalization support for OSCAL JSON documents

**Branch:** `feat/json-canonicalization-scheme-2013`
Closes #2013 

---

## Problem

OSCAL documents are represented as JSON, but JSON has no canonical form.
The same logical document can be serialized in countless ways depending on key order, whitespace, and float formatting.
This makes it impossible to produce a stable, reproducible byte sequence to feed into a cryptographic hash or signature without a canonicalization step.

[RFC 8785 (JSON Canonicalization Scheme, JCS)](https://www.rfc-editor.org/rfc/rfc8785) defines exactly this: a deterministic, compact JSON form where:
- Object keys are sorted by their **UTF-16 code-unit** byte sequence
- Numbers are normalized (integer-valued floats lose the decimal point)
- Strings have minimal and fully specified escaping
- There is no insignificant whitespace

This PR implements the three items listed in issue #2013:

> 1. Introduce RFC 8785 formatted JSON (aka JCS) as a supported output type.
> 2. Introduce flags throughout CLI commands to support this.
> 3. Introduce a command to allow object canonicalization.

Caveats from the issue are fully respected: YAML and XML are out of scope, the third-party `jcs` library is not used.

---

### 1. `trestle/common/jcs.py` pure-stdlib RFC 8785 implementation

No new runtime dependency.  A single `canonicalize(data) -> bytes` function that:

- Handles all JSON types (`None`, `bool`, `int`, `float`, `str`, `list`, `dict`)
- Sorts dict keys by **UTF-16BE** code-unit sequence (not UTF-8, not Unicode code-point order) which is what §3.2.3 of the spec mandates
- Normalizes integer-valued floats (`1.0 → "1"`, `-0.0 → "0"`)
- Non-finite floats (`NaN`, `Infinity`) raise `ValueError` as required by the spec
- Control characters `< U+0020` are always `\uXXXX`-escaped

### 2. `OscalBaseModel.oscal_serialize_jcs()` new method on the base model

```python
# Get JCS bytes for any OSCAL object
catalog.oscal_serialize_jcs()           # wrapped: {"catalog": {...}}
catalog.oscal_serialize_jcs(wrapped=False)  # unwrapped
```

Internally serializes via the existing `orjson` pipeline first (so `datetime` → ISO-8601 and all other pydantic encoders are applied correctly), then re-parses and canonicalizes.  This is the correct place to hook canonicalization into the object model.

### 3. `trestle canonicalize`: new CLI command

```bash
# Write canonical JSON to stdout (suitable for piping to sha256sum / openssl)
trestle canonicalize -f catalogs/mycatalog/catalog.json

# Write canonical JSON to a file
trestle canonicalize -f catalogs/mycatalog/catalog.json -o canonical.json
```

- Accepts any OSCAL JSON file in the trestle workspace (file path relative to trestle root)
- Handles both monolithic files and split/distributed models (`load_distributed`)
- Rejects `.yaml` / `.yml` with a clear error referencing the scope decision in #2013
- Registered as a first-class subcommand of `trestle` in `cli.py`

### 4. `--jcs` / `-j` flag on `trestle import`, `trestle assemble`, `trestle create`

The flag causes the written JSON output to be the JCS canonical form instead of the default pretty-printed form.

```bash
# Import a catalog and store it in canonical form
trestle import -f external/catalog.json -o mycat --jcs

# Assemble a split model into the dist directory in canonical form  
trestle assemble catalog -n mycat --jcs

# Create a new skeleton model in canonical form
trestle create -t catalog -o newcat --jcs
```

All three commands validate that `--jcs` is only used with JSON output (`--extension json` or `.json` path) and raise a `TrestleError` otherwise.

---

## Files changed

| File | Change |
|---|---|
| `trestle/common/jcs.py` | **New** : pure-stdlib RFC 8785 implementation |
| `trestle/core/base_model.py` | Add `oscal_serialize_jcs()` to `OscalBaseModel`; import `canonicalize` |
| `trestle/core/commands/canonicalize.py` | **New** : `trestle canonicalize` command |
| `trestle/cli.py` | Register `CanonicalizeCmd` |
| `trestle/core/commands/import_.py` | Add `--jcs` / `-j` flag |
| `trestle/core/commands/assemble.py` | Add `--jcs` / `-j` flag |
| `trestle/core/commands/create.py` | Add `--jcs` / `-j` flag |
| `tests/trestle/core/jcs_test.py` | **New** : 16 unit tests for the RFC 8785 implementation |
| `tests/trestle/core/commands/canonicalize_test.py` | **New** : 7 integration tests for the command and `--jcs` flag |

---

## Completion criteria coverage

| Criterion from issue | Status |
|---|---|
| Introduce RFC 8785 / JCS as a supported output type | ✅ `oscal_serialize_jcs()` on every OSCAL model |
| Introduce flags throughout CLI commands | ✅ `--jcs` on `import`, `assemble`, `create` |
| Introduce a command to allow object canonicalization | ✅ `trestle canonicalize` |
| YAML is out of scope | ✅ rejected with clear error |
| XML is out of scope | ✅ not applicable (trestle works with JSON/YAML only) |
| `jcs` library is out of scope | ✅ pure stdlib (`json`, `math`) |
| Purely canonicalization : no reproducible-build scope creep | ✅ |

---

## Design decisions

**Why no new runtime dependency?**
RFC 8785 is small and well-specified. The key algorithm is sorting by UTF-16BE bytes and normalizing numbers. Shipping a pure-stdlib implementation avoids adding a dependency whose maintenance lifecycle we can't control, and makes it trivially auditable.

**Why serialize via orjson first, then canonicalize?**
Pydantic v1 uses custom encoders (e.g. `datetime` → ISO-8601 string) that are wired into `orjson`. Re-implementing all those encoders inside `canonicalize()` would be fragile.  The two-step approach (serialize → re-parse → canonicalize) is slightly less efficient but is correct and will stay correct as encoders evolve.

**Why is `--jcs` on `import`/`assemble`/`create` but not `split`/`merge`?**
`split` and `merge` work with partial model fragments, not complete top-level documents. A canonical form for a fragment is meaningless for signing (you sign the whole document). The three commands that write complete top-level OSCAL documents received the flag.

---

## Testing

```
tests/trestle/core/jcs_test.py       : 16 unit tests  (all pass)
tests/trestle/core/commands/canonicalize_test.py : 7 integration tests (all pass)
```

Unit tests cover: null, bool, int, float (including integer-valued and non-integer), NaN/Infinity rejection, string escaping (all control chars, backslash, quote), Unicode passthrough, empty/non-empty list, dict key sorting (ASCII and non-ASCII UTF-16 ordering), nested structures, determinism, and unsupported-type rejection.

Integration tests cover: stdout output, file output, determinism across two runs, key-sort property on a live OSCAL catalog, nonexistent-file error, YAML rejection, and `trestle import --jcs`.

---

## Checklist

- [x] All three completion criteria from issue #2013 implemented
- [x] No new runtime dependencies
- [x] YAML and XML correctly out of scope
- [x] `jcs` library not used
- [x] 23 tests : 23 passing
- [x] `ruff` lint clean on all changed files
- [x] DCO `Signed-off-by` present
- [x] Conventional commit message: `feat: add RFC 8785 (JCS) canonicalization support for OSCAL JSON documents`

### DIFF
--- a/tests/trestle/core/commands/canonicalize_test.py
+++ b/tests/trestle/core/commands/canonicalize_test.py
@@ -1,0 +1,166 @@
+# -*- mode:python; coding:utf-8 -*-
+
+# Copyright (c) 2020 IBM Corp. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Tests for trestle canonicalize command."""
+
+import json
+import pathlib
+import secrets
+import string
+import sys
+
+from _pytest.monkeypatch import MonkeyPatch
+
+import pytest
+
+from tests.test_utils import execute_command_and_assert
+
+from trestle.cli import Trestle
+from trestle.core import generators
+from trestle.oscal.catalog import Catalog
+
+
+def _write_catalog(tmp_dir: pathlib.Path) -> pathlib.Path:
+    """Write a generated catalog JSON to a temp file outside the trestle workspace."""
+    rand_str = ''.join(secrets.choice(string.ascii_letters) for _ in range(16))
+    catalog_file = tmp_dir.parent / f'{rand_str}.json'
+    catalog_data = generators.generate_sample_model(Catalog)
+    catalog_data.oscal_write(catalog_file)
+    return catalog_file
+
+
+def test_canonicalize_to_stdout(
+    tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPatch, capsys
+) -> None:
+    """Canonicalize command writes valid JSON to stdout when no -o is given."""
+    # First import a catalog so it exists in the workspace
+    catalog_file = _write_catalog(tmp_trestle_dir)
+    execute_command_and_assert(f'trestle import -f {catalog_file} -o mycat', 0, monkeypatch)
+
+    catalog_path = tmp_trestle_dir / 'catalogs' / 'mycat' / 'catalog.json'
+    assert catalog_path.exists()
+
+    testcmd = f'trestle canonicalize -f catalogs/mycat/catalog.json'
+    monkeypatch.setattr(sys, 'argv', testcmd.split())
+    rc = Trestle().run()
+    assert rc == 0
+
+    captured = capsys.readouterr()
+    # Output must be valid JSON with no leading/trailing whitespace lines
+    output = captured.out.strip()
+    parsed = json.loads(output)
+    assert 'catalog' in parsed
+
+
+def test_canonicalize_to_file(
+    tmp_trestle_dir: pathlib.Path, monkeypatch: MonkeyPatch
+) -> None:
+    """Canonicalize command writes canonical JSON to file when -o is given."""
+    catalog_file = _write_catalog(tmp_trestle_dir)
+    execute_command_and_assert(f'trestle import -f {catalog_file} -o mycat2', 0, monkeypatch)
+
+    out_file = tmp_trestle_dir / 'canonical.json'
+    testcmd = f'trestle canonicalize -f catalogs/mycat2/catalog.json -o canonical.json'
+    monkeypatch.setattr(sys, 'argv', testcmd.split())
+    rc = Trestle().run()
+    assert rc == 0
+
+    assert out_file.exists()
+    content = json.loads(out_file.read_bytes())
+    assert 'catalog' in content
+
+
+def test_canonicalize_is_deterministic(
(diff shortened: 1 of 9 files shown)


degenaro:
Fixed by #2227

Issue #2089, opened 2026-02-20
Duplicate UUIDs in component definition test data files


LAKSHJAIN14:
## Description

Several component definition test data files in tests/data/json/ contain duplicate control-implementation UUIDs within the same document, which violates the OSCAL specification requirement that all UUIDs must be unique within a document.
In each case, two different components within the same file share the same UUID for their control-implementation objects. This appears to be a copy-paste oversight when the test data was originally created.

for eg. comp_def_a.json, comp_def_b.json

## Current behaviour

- trestle import correctly rejects these files via DuplicatesValidator in duplicates_validator.py, which checks for unique UUIDs using ModelUtils.has_no_duplicate_values_by_name(model, 'uuid').

- However, tests using test_utils.load_from_json bypass this validation entirely, so tests using these files pass despite the invalid data.

## Questions for discussion

1.Should the duplicate UUIDs in the test data files simply be fixed (assign unique UUIDs to each control-implementation)?

2.Should other commands (e.g., ssp-generate, component-generate) also validate for duplicate UUIDs when loading component definitions, or is the import-time check sufficient?

 @degenaro @vikas-agarwal76

vikas-agarwal76:
@degenaro We need a discussion on this.

PR #2142, opened 2026-03-11
fix: duplicate uuid issue in test data by adding a new load_valid_model_from_json function

LAKSHJAIN14:
## Types of changes

## Summary

As discussed during the community meeting, This is the first PR related to the issue #2089 which fixes all the duplicate uuids in comp_def test files that were there and I have looked if there exists a current check which checks for duplicate uuid and yeah there is a check in /tests/trestle/core/commands/validate_test.py but it doesn't load any of the file from the test data , it loads comp_def using the sample_component_definition pytest fixture and then changes uuid to become duplicate and then check which is sufficent I guess,

There is another file with duplicate uuid which is tests/data/yaml/bad_component_dup_uuid.yaml but I couldn't find any test function using it so i think its a dead data and this file is also there in the issue #2116 ,So I guess we can handle it there.

If these all things seem ok then I could look into the test_utils.load_from_json function and have it also validate like it happens when we do trestle import but I think we could do that in another PR and discuss it there.

@degenaro @vikas-agarwal76


degenaro:
@LAKSHJAIN14 Please resolve conflicts, thx!

LAKSHJAIN14:
resolved the conflicts, PTAL
@degenaro

PR #2249, opened 2026-05-29
feat: pydantic v2 mode

degenaro:
## Types of changes

- [ ] My code follows the code style of this project.
- [ ] My PR meets testing requirements.
- [ ] All new and existing tests passed.

## Summary

Drops the pydantic.v1 compatibility shim used throughout the codebase and migrates every import, API call, and model configuration to the native Pydantic v2 surface. The generated OSCAL models under trestle/oscal/ are regenerated with datamodel-code-generator's --output-model-type pydantic_v2.BaseModel flag. Test fixtures, CSV data, and JSON test files are updated to match the new validation rules (e.g. lowercase component type values, removal of empty alters: [] fields).

## Key links:

# Before you merge


degenaro:
Bob's review:

## PR #2249 — `feat: pydantic v2 mode`
**Repo:** oscal-compass/compliance-trestle | **Base:** `develop` ← **Head:** `pydantic-v2-mode`
**State:** OPEN | +5,427 / -3,104 lines | 127 files changed | 0 reviews so far

---

### Purpose & Scope

Drops the `pydantic.v1` compatibility shim used throughout the codebase and migrates every import, API call, and model configuration to the native **Pydantic v2** surface. The generated OSCAL models under `trestle/oscal/` are regenerated with `datamodel-code-generator`'s `--output-model-type pydantic_v2.BaseModel` flag. Test fixtures, CSV data, and JSON test files are updated to match the new validation rules (e.g. lowercase component `type` values, removal of empty `alters: []` fields).

This is a large, well-scoped migration PR. The vast majority of the diff is mechanical (`import` changes, `.v1` removal, `__root__` → `root`, `parse_obj` → `model_validate`). Substantive logic changes are concentrated in five files reviewed below.

---

### Change Structure

| Category | Files | Nature |
|---|---|---|
| Core model layer | `base_model.py`, `type_utils.py`, `model_utils.py` | Substantive rewrites |
| Generator | `generators.py` | Major refactor + new helpers |
| Element/plan | `elements.py`, `plans.py`, `actions.py` | Logic changes |
| OSCAL codegen | `oscal_normalize.py`, `gen_oscal.py` | v2 codegen support |
| Supporting utilities | `str_utils.py`, `list_utils.py`, `generic_oscal.py`, `parser.py` | Mechanical + minor logic |
| Test fixtures & data | ~75 JSON/CSV + ~20 test files | Data normalization + new tests |
| Build tooling | `Makefile`, `pyproject.toml` | hatch / coverage config |

---

### `trestle/core/base_model.py` — Core Migration

**✅ Good** — Correct migration of the inner `class Config` to `model_config = ConfigDict(...)` with equivalent settings: `populate_by_name=True`, `extra='forbid'`, `validate_assignment=True`.

**⚠️ Concern — datetime serialization approach**
The `@field_serializer('*', mode='wrap', when_used='json')` wildcard serializer is applied to *every* field on every OSCAL model. Its only real job is to (a) stringify `AnyUrl` and (b) replace trailing `Z` with `+00:00` on datetime strings. This is expensive and fragile:
- The `'Z' → '+00:00'` string-replace fires on *any* serialized string ending with `Z`. It's character-level, not semantically aware.
- A targeted serializer on `datetime` alone (or `json_encoders={datetime: robust_datetime_serialization}`, still valid in v2, or a `PlainSerializer` annotation) would be cleaner and faster.
- The comment inside `model_config` saying *"json_loads is not directly configurable in ConfigDict"* is inaccurate — Pydantic v2 does support `model_config['json_loads']`.

**⚠️ Concern — `__eq__` override uses `model_dump()`**
The new `__eq__` calls `model_dump()` on both sides for every equality check. `model_dump()` serializes the entire object graph and is not free. This will regress performance in any code that compares models in a loop (plan diffing, control resolution). Consider comparing `__dict__` directly or at least short-circuiting on `model_fields_set` before falling back to `model_dump()`.

**⚠️ Info — silent datetime output format change**
`robust_datetime_serialization` now omits sub-second precision when microseconds are zero (`timespec='seconds'` instead of `timespec='milliseconds'`). Both forms are valid OSCAL, but this silently changes the output format for stored datetimes (e.g. `2024-01-01T00:00:00.000+00:00` → `2024-01-01T00:00:00+00:00`). Downstream consumers that do exact string comparison will be affected. This intentional change should be documented in the PR description.

**🔴 Issue — unused `orjson` import**
`orjson` is still imported at the top of `base_model.py` but is no longer wired into `model_config` (the `json_loads = orjson.loads` line was removed without a replacement). Either use it — `json_loads=orjson.loads` is still valid in v2's `ConfigDict` — or remove the import.

**ℹ️ Note — `FieldWrapper` lifetime**
`FieldWrapper` delegates `FieldInfo` attribute access via `__getattr__` to preserve call-sites that used to access `ModelField` attributes. Reasonable shim, but it needs a `# TODO` listing which call-sites still depend on it so it can be removed once they are updated.

---

### `trestle/core/generators.py`

**✅ Good** — Clean decomposition of the monolithic `generate_sample_value_by_type` into `_handle_enum_types`, `_handle_simple_types`, `_handle_constrained_string`, `_handle_special_types`, etc. The helpers are individually unit-testable.

**🔴 Issue — duplicate constraint-extraction logic**
`_extract_int_constraints` and `_get_constrained_int_value` both iterate over constraint metadata and compute identical `floor` / `multiple_of` logic. One is superfluous — `_get_constrained_int_value` should call `_extract_int_constraints` internally rather than re-implementing it.

**⚠️ Concern — `_is_constrained_string` heuristic**
The check `'constr' in str(type_).lower()` matches the type's string repr. This can produce false positives for any type whose repr happens to contain `constr`. Prefer `typing.get_origin(type_) is Annotated and str in typing.get_args(type_)` or check against the actual `pydantic.types` constraint classes directly.

**⚠️ Concern — `constr` is deprecated in Pydantic v2**
`constr` is imported and used in `generators.py` and `generic_oscal.py`. In Pydantic v2, `constr` is a compatibility alias for `Annotated[str, StringConstraints(...)]` — it still works but emits deprecation warnings and will be removed in a future release. Should be migrated or at minimum tracked as a known TODO.

---

### `trestle/common/model_utils.py`

**✅ Good** — `_get_model_type_from_union` now has a correct two-phase lookup: alias match first, then field-map match. The fallback that skips `list`-origin Union arms is a correct fix.

**⚠️ Concern — inline imports inside `_get_primary_model_instance`**
`import json` and `import yaml` are deferred inside the method body. `yaml` is not in the module-level imports. Move both to the top of the file to avoid repeated import cost and to follow project conventions.

**⚠️ Concern — English pluralisation heuristics in `_resolve_collection_item_alias`**
The new fallbacks (`endswith('ies')` → strip to `y`, `endswith('s')` → strip `s`) are English-language assumptions about OSCAL field naming. They match current conventions but are fragile if new fields are added. Add a comment documenting the assumption and a test covering each branch.

---

### `trestle/common/type_utils.py`

**🔴 Issue — magic string comparison for `types.UnionType`**
Multiple places compare `str(origin) == "<class 'types.UnionType'>"` (via the `_UNION_TYPE_STR` constant) to detect Python 3.10+ `X | Y` union syntax. This is fragile — the string representation is CPython-implementation-specific. Replace with:

```python
import types
# Python 3.10+
isinstance(origin, types.UnionType)
# Or portable:
origin is Union or (hasattr(types, 'UnionType') and isinstance(origin, types.UnionType))
```

The same string-compare is also duplicated verbatim in `elements.py` — both should be fixed together.

**⚠️ Concern — `_get_root_type_name` capitalises origin name**
`origin.__name__.capitalize()` returns `'List'` for `list`, which callers then check against `root_type == 'List'`. This is a case-sensitive string comparison that depends on capitalisation behaviour of `__name__`. Prefer comparing the origin directly (`origin is list`, `origin is dict`) rather than string-matching on names.

---

### `trestle/core/models/elements.py`

**✅ Good** — The restructured `get_type()` control flow (collection branch first, then alias field-map lookup, then union alias match) is clearer than the previous version and correctly handles `Optional[List[T]]` traversal paths like `catalog.controls.control`.

**⚠️ Concern — redundant inline import**
`import typing_extensions` appears inside the body of `get_type()` at the tail of the method, even though `typing_extensions` is already imported at the top of the file. Remove it.

**⚠️ Concern — duplicated `UnionType` magic-string check**
Same issue as in `type_utils.py` — `str(origin) == "<class 'types.UnionType'>"` appears again at the bottom of `get_type()`. Consolidate into a single helper in `type_utils.py` and call it from both places.

---

### `scripts/oscal_normalize.py` (codegen script)

**✅ Good** — Removing `pydantic_interface_v1()` and switching from `.update_forward_refs()` to `model_rebuild()` is correct. Disabling the custom `__get_validators__` union validators (which caused infinite recursion in v2) and relying on native Union dispatch is the right call.

**⚠️ Concern — `add_union_validators` is dead code**
The function now unconditionally returns `''`. Its call-sites still invoke it, write the empty string, and then perform `seen_group1`/`seen_group2` tracking work around it. Either delete the function entirely or consolidate the call-sites so the tracking logic doesn't depend on a no-op call. Leaving a permanently-empty function creates confusion for future maintainers.

**ℹ️ Note — `find_forward_refs` heuristic**
The new implementation searches for `' = Union['` text patterns in generated code lines. This is codegen-only logic, but it is sensitive to whitespace/formatting. A comment should make clear this function only operates on code generated by this script and should never be used at runtime.

---

### Supporting Changes

| File | Change | Assessment |
|---|---|---|
| `str_utils.py` | `__root__` → `root` in `string_from_root`; `root_suffix_aliases` dict in `classname_to_alias`; digit-strip moved inside the `_camel_to_dash` call | ✅ Fixes alias for `Group1` → `group` correctly |
| `list_utils.py` | `none_if_empty` parameter widened to `Optional[List[TG]]`; improved docstrings for Pydantic v2 `min_length` pattern | ✅ Good |
| `generic_oscal.py` | `regex=` → `pattern=` in all `constr()` calls; `none_if_empty` applied to `by_comps` | ✅ Necessary for Pydantic v2 |
| `parser.py` | `parse_obj(data)` → `model_validate(data)` | ✅ Correct |
| `plans.py` | `__eq__` expanded to iterate and compare actions; new `__repr__` | ⚠️ `strict=True` on `zip()` is redundant since length is already checked above |
| Test data JSON | Lowercase `"type": "service"`; `"combine": {}`; removed `"alters": []` | ⚠️ Silent data-model normalisations — should be explained in the PR description |
| `pyproject.toml` | Parallel coverage config; removed `warn_untyped_fields`; tightened `disable_error_code` list | ✅ Positive — tightening mypy is good |

---

### Issues Summary

| # | Severity | Location | Description |
|---|---|---|---|
| 1 | 🔴 Issue | `base_model.py` | `orjson` imported but not used — wire it in or remove it |
| 2 | 🔴 Issue | `generators.py` | `_extract_int_constraints` and `_get_constrained_int_value` duplicate identical logic |
| 3 | 🔴 Issue | `type_utils.py`, `elements.py` | Magic string `UnionType` detection — use `isinstance(origin, types.UnionType)` |
| 4 | ⚠️ Warn | `base_model.py` | Wildcard `@field_serializer('*')` is expensive — use a targeted datetime serializer |
| 5 | ⚠️ Warn | `base_model.py` | `__eq__` via `model_dump()` is O(n) deep serialization on every comparison |
| 6 | ⚠️ Warn | `base_model.py` | Silent datetime output format change (zero-microsecond → seconds timespec) |
| 7 | ⚠️ Warn | `generators.py` | `_is_constrained_string` uses fragile repr-string matching |
| 8 | ⚠️ Warn | `model_utils.py` | Inline `import json` / `import yaml` inside method body |
| 9 | ⚠️ Warn | `model_utils.py` | English pluralisation heuristics need tests and a comment |
| 10 | ⚠️ Warn | `type_utils.py` | `_get_root_type_name` relies on string capitalisation of `origin.__name__` |
| 11 | ⚠️ Warn | `elements.py` | Redundant `import typing_extensions` inside method body |
| 12 | ⚠️ Warn | `oscal_normalize.py` | `add_union_validators` is dead code (always returns `''`) but call-sites remain |
| 13 | ⚠️ Warn | Multiple | `constr` is a deprecated Pydantic v2 alias — migrate to `Annotated[str, StringConstraints(...)]` |
| 14 | ℹ️ Info | `base_model.py` | `FieldWrapper` needs a lifetime TODO to track call-site migration |
| 15 | ℹ️ Info | PR description | Summary section is empty — no explanation of data model changes or migration rationale |

---

### Verdict

> **🟡 Request Changes (minor) before merge**

The migration is technically correct and covers the full breadth of the codebase. All the mechanical changes are right. The new test coverage is valuable.

**The three 🔴 Issues (#1–3) should be fixed before merge.** The ⚠️ Warnings are non-blocking but items 4, 5, and 12 in particular (wildcard serializer, `model_dump()` in `__eq__`, dead `add_union_validators`) should be tracked in follow-up issues before more call-sites accumulate. The PR description Summary should be filled in to document the intentional data-model normalisations.

degenaro:
coverage: 98%

tests/trestle/core/generator_test.py
```diff
@@ -91,7 +88,7 @@ def test_generate_sample_model() -> None:
             'title': const.REPLACE_ME,
             'last-modified': '2020-10-21T06:52:10.387+00:00',
             'version': const.REPLACE_ME,
-            'oscal-version': oscal.OSCAL_VERSION,
+            'oscal-version': const.REPLACE_ME,
         },
```
vikas-agarwal76, line 92: Why is oscal version not used?

tests/trestle/core/profile_resolver_test.py
```diff
@@ -171,10 +172,9 @@ def test_all_positions_for_alter_can_be_resolved(tmp_trestle_dir: pathlib.Path)
...
     """Test profile resolver merge."""
     profile = gens.generate_sample_model(prof.Profile)
-    method = 'merge'
-    # In OSCAL 1.2.0, combine is a dict, not a Combine object
-    # Merge1 has combine as dict and flat as required dict
-    profile.merge = prof.Merge1(combine={'method': method}, flat={})
+    # In OSCAL 1.2.0, combine is an empty dict per schema (additionalProperties: false)
+    # Merge1 has combine as empty dict and flat as required empty dict
+    profile.merge = prof.Merge1(combine={}, flat={})
     merge = Merge(profile)
```
vikas-agarwal76, line 178: Combine can have a method.

tests/trestle/core/profile_resolver_test.py
```diff
@@ -212,15 +212,16 @@ def test_profile_resolver_merge(sample_catalog_rich_controls: cat.Catalog) -> No
...
 
-    # add part to first control and merge but with use-first.  The part should not be there at end.
-    method = prof.CombinationMethodValidValues.use_first.value
-    # In OSCAL 1.2.0, combine is a dict, not a Combine object
-    profile.merge = prof.Merge1(combine={'method': method}, flat={})
+    # add part to first control and merge. Since combine is empty per OSCAL schema,
+    # the default merge behavior applies and both parts are included.
+    # In OSCAL 1.2.0, combine is an empty dict per schema (additionalProperties: false)
+    profile.merge = prof.Merge1(combine={}, flat={})
     merge = Merge(profile)
```
vikas-agarwal76, line 219: combine can have a method.

tests/trestle/tasks/csv_to_oscal_cd_test.py
```diff
@@ -1510,7 +1528,7 @@ def test_execute_validation(tmp_path: pathlib.Path) -> None:
     assert component.props[1].value == 'validation-check-id'
     assert component.props[2].name == 'Check_Description'
     assert component.props[2].value == 'validation-check-description'
-    assert len(component.control_implementations) == 0
+    assert component.control_implementations is None or len(component.control_implementations) == 0
```
vikas-agarwal76, line 1532: Why OR here as lists cannot be empty, so only none check should be used.

tests/trestle/transforms/transformer_factory_test.py
```diff
@@ -31,29 +31,33 @@ class DummyTransformer(TransformerBase):
...
 
 
 class DummyResultsTransformer(ResultsTransformer):
     """Example results transformer."""
 
     def transform(self, blob: str) -> Results:
         """Transform the blob."""
-        return Results(__root__=[])
+        # Pydantic v2: RootModel uses 'root' instead of '__root__'
+        return Results(root=[])
```
vikas-agarwal76, line 58: Are empty list allowed?

tests/trestle/transforms/transformer_factory_test.py
```diff
@@ -31,29 +31,33 @@ class DummyTransformer(TransformerBase):
...
+    assert transformer.transform('foo') == RoleId(root='my_string')
 
 
 def test_results_transformer() -> None:
     """Test results transformer."""
     tf.register_transformer('dummy', DummyResultsTransformer)
     transformer: ResultsTransformer = tf.get('dummy')
-    assert transformer.transform('foo') == Results(__root__=[])
+    # Pydantic v2: RootModel uses 'root' instead of '__root__'
+    assert transformer.transform('foo') == Results(root=[])
```
vikas-agarwal76, line 73: Same, are empty list allowed?

vikas-agarwal76 (commented):
Some commnets

tests/trestle/core/generator_test.py
degenaro: fixed.

degenaro:
# OSCAL 1.0.4 → 1.2.1: Breaking Changes Analysis

> Trestle Pydantic model diff — what breaks when loading pre-1.2.0 OSCAL documents

**Key finding:** The OSCAL version string validator is **disabled** (`strict_version = False` in `trestle/oscal/common.py`), so any version string loads without error. However, the **Pydantic model structure** has changed extensively — older documents will fail at parse time with cryptic validation errors, not a version error.

---

## 1. Profile — Structural Breaks

| Change | 1.0.4 shape | 1.2.1 shape | Impact |
|---|---|---|---|
| `merge.combine` | `Combine` object with `.method` attribute | `dict[str, Any]`, `max_length=0` (empty dict only) | 🔴 **BREAKS** — any `"combine": {"method": "..."}` fails |
| `merge` type | Single `Merge` class | `Union[Merge1, Merge2, Merge3]` discriminated by required field | 🔴 **BREAKS** — old documents with multiple merge fields fail `additionalProperties: false` |
| `import` type | Single `Import` class | `Union[Import1, Import2]` | 🟡 Changed — likely compatible |
| `group` type | Single `Group` class | `Union[Group1, Group2]` | 🟡 Changed — likely compatible |
| `set_parameters` type | `List[SetParameter]` | `list[SetParameters \| SetParameters1]` | 🟡 Changed — renamed + split into union |
| `Remove.by_item_name` | Free-form `constr(regex=...)` | `ItemNameValidValues` (enum) | 🔴 **BREAKS** — values not in the enum fail |
| Classes removed | `Combine`, `Config`, `Group`, `Import`, `InsertControls`, `Matching`, `Merge`, `Method`, `Order`, `Position`, `SelectControlById`, `SetParameter`, `WithChildControls`, `WithId` | — | Python API break for code using these directly |

---

## 2. Common — Structural Breaks

These affect **all** OSCAL document types (catalog, SSP, profile, component, AR, AP, POAM).

| Change | 1.0.4 shape | 1.2.1 shape | Impact |
|---|---|---|---|
| `Metadata.last_modified` | `datetime` (naive allowed) | `AwareDatetime` (timezone required) | 🔴 **BREAKS** — datetimes without timezone offset fail |
| `Metadata.published` | `Optional[datetime]` | `AwareDatetime \| None` | 🔴 **BREAKS** — naive datetime fails |
| `Metadata.parties` | `List[Party]` | `list[Parties \| Parties1]` | 🟡 Likely breaks — `Party` class removed |
| `AssessmentSubject` | Flat fields (`type`, `description`, etc.) | `root: AssessmentSubjectAll \| AssessmentSubjectSpecific` (RootModel union) | 🔴 **BREAKS** — completely restructured |
| `Base64.value` | Plain `str` | `constr(pattern='^[0-9A-Za-z+/]+={0,2}$')` | 🔴 **BREAKS** — non-base64 strings fail |
| `Property.group` | Not present | Added | 🟢 Additive — compatible |
| `Link.resource_fragment` | Not present | Added | 🟢 Additive — compatible |
| `Location.address` | Required | Optional | 🟢 Relaxed — compatible |
| RootModel classes | `__root__: X` (Pydantic v1) | `root: X` (Pydantic v2) | 🟢 JSON-transparent — no data impact |
| Classes removed | `ControlObjectiveSelection`, `HowMany`, `MemberOfOrganization`, `Parameter`, `Party`, `RelatedRisk`, `StatementId`, `Timing`, `Transport`, `Type1/2/3`, `Unit`, `Url`, `Value` | — | Python API break |

---

## 3. Summary by Severity

| Severity | Change | Affected types |
|---|---|---|
| 🔴 Hard break | `merge.combine` must be empty dict — no `method` key allowed | Profile |
| 🔴 Hard break | Datetime fields require timezone offset (`AwareDatetime`) | **All** |
| 🔴 Hard break | `AssessmentSubject` restructured into discriminated union | AR, AP |
| 🔴 Hard break | `Merge` split into `Union[Merge1, Merge2, Merge3]` with `additionalProperties: false` | Profile |
| 🔴 Hard break | `Remove.by_item_name` is now an enum | Profile |
| 🟡 Likely break | `Metadata.parties` changed from `Party` to `Parties \| Parties1` | **All** |
| 🟡 Likely break | Classes renamed with numbering (`Group→Group1/2`, `Import→Import1/2`, etc.) | Profile, Catalog |
| 🟢 Compatible | Additive fields (`Property.group`, `Link.resource_fragment`, `Remove.remarks`) | All |
| 🟢 Compatible | `Optional[X]` → `X \| None` syntax only | All |

---

## 4. Conclusion

Pre-1.2.0 documents fail on at least **4 distinct hard-break categories** — most universally the `AwareDatetime` requirement (affects every document type) and the `Merge` structural split.

With `strict_version = False`, users get cryptic Pydantic field errors with no indication of the root cause. Enabling `strict_version = True` in [`trestle/oscal/common.py`](trestle/oscal/common.py) is the minimal, honest fix — it surfaces a clear `"OSCAL version X.Y.Z is not supported"` message instead.

tests/trestle/core/profile_resolver_test.py
degenaro: @vikas-agarwal76 The schema does not seem to have method, though the online doc does?

image

tests/trestle/tasks/csv_to_oscal_cd_test.py
degenaro: fixed.

tests/trestle/transforms/transformer_factory_test.py
degenaro: fixed. there are several cases. for model construct, empty list is allowed. but for read/write model empty list is disallowed.

# `Results` Empty-List Validation Behaviour

The [`Results`](../trestle/transforms/results.py) class mirrors the OSCAL JSON Schema
constraint `minItems: 1` on the `results` array via `Field(min_length=1)`.
The table below shows what is allowed or rejected at each operation.

| Operation | Empty list allowed? | Enforced by |
|---|---|---|
| `Results.model_construct(root=[])` | ✅ Yes | — intentional; bypasses Pydantic validation for the builder pattern used by all transformer implementations (`xccdf`, `osco`, `tanium`). Append at least one `Result` before writing. |
| `Results(root=[])` | ❌ No | Pydantic `min_length=1` on `root` field |
| `Results.model_validate([])` — called by `oscal_read` | ❌ No | Pydantic `min_length=1` on `root` field |
| `results.oscal_write(path)` with empty list | ❌ No | Explicit guard in `Results.oscal_write()` raises `TrestleError` before any file is written |

## Why `model_construct` is needed for the builder pattern

All transformer implementations follow this pattern:

```python
results = Results.model_construct(root=[])   # empty — validation skipped
results.root.append(self._results_factory.result)
return results
```

Using `Results(root=[])` (direct construction) would raise a `ValidationError` immediately
because Pydantic v2 validates field constraints — including `min_length=1` — at construction
time, before any items can be appended.

`model_construct` is the correct Pydantic v2 mechanism for building an object
incrementally. The `oscal_write` guard ensures a `Results` that was never populated
cannot be persisted as an invalid OSCAL document.

## Schema reference

[`release-schemas/oscal_assessment-results_schema.json`](../release-schemas/oscal_assessment-results_schema.json):

```json
"results": {
  "type": "array",
  "minItems": 1,
  ...
}
```

tests/trestle/transforms/transformer_factory_test.py
degenaro: fixed. added missing test cases as well.

tests/trestle/core/profile_resolver_test.py
vikas-agarwal76: Since combine is a dict, we can still use the old combine with a non-empty dict. Any reason for making it now?

tests/trestle/core/profile_resolver_test.py
vikas-agarwal76: Same as previous comment. Combine can still be a dict, with method and is not required to be empty.

tests/trestle/core/profile_resolver_test.py
degenaro: @vikas-agarwal76 Restoring test code back to original failed. The code is correct as is. Here's why:

Here is the definitive evidence from the OSCAL JSON Schema in [release-schemas/oscal_profile_schema.json](vscode-webview://10e5boks6pch0rrt19dlrd3tpau6p5c4s7v32fkk38pc0ev4jl4l/release-schemas/oscal_profile_schema.json):

flat (lines 94–98)
"flat" : {
  "title" : "Flat Without Grouping",
  "description" : "Directs that controls appear without any grouping structure.",
  "type" : "object",
  "additionalProperties" : false
}

additionalProperties: false with no properties defined → this is a JSON Schema empty-object type. It cannot have any properties at all. The max_length=0 in the Python model correctly reflects this — flat must always be {}.

combine (lines 89–93, 104–108, 118–122)
"combine" : {
  "title" : "Combination Rule",
  "description" : "A Combine element defines how to resolve duplicate instances of the same control (e.g., controls with the same ID).",
  "type" : "object",
  "additionalProperties" : false
}

Identical structure to flat — additionalProperties: false, no properties defined. According to the OSCAL schema, combine is also supposed to be an empty object {} — it cannot hold any keys either.

Conclusion
The OSCAL 1.2.0 JSON schema says both combine and flat must be empty objects (additionalProperties: false with no properties declared). The max_length=0 on both fields in the Python model is correct per the schema.

vikas-agarwal76: So it means that in pydantic v1 this empty combine was not enforced but now it is, right? and hence the test need to change.

tests/data/csv/bp.sample.v2.csv
```diff
@@ -1,12 +1,12 @@
...
-3000032,iam_admin_role_serviceid_maxcount,Ensure IAM-enabled services have no more than # service IDs with the IAM administrator role,iam_admin_role_serviceid_maxcount,Check whether IAM-enabled services have no more than # service IDs with the IAM administrator role,,,https://github.com/usnistgov/oscal-content/blob/main/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_HIGH-baseline_profile.json,NIST Special Publication 800-53 Revision 5 HIGH IMPACT BASELINE,Service,ac-6 ac-5_smt.c ia-7,IAM,IAM,no_of_service_id_admins_for_iam,Maximum no of IAM Service ID,3,3,http://scal-compass.github.io/compliance-trestle/schemas/oscal/cd
+3000020,account_owner_authorized_ip_range_configured,Ensure authorized IP ranges are configured by the account owner,account_owner_authorized_ip_range_configured,Check whether authorized IP ranges are configured by the account owner,,,https://github.com/usnistgov/oscal-content/blob/main/nist.gov/SP800-53/rev5/json/NIST_SP-800-53_rev5_HIGH-baseline_profile.json,NIST Special Publication 800-53 Revision 5 HIGH IMPACT BASELINE,service,sc-7_smt.a sc-7_smt.b sc-7.3 sc-7.4_smt.a sc-7.5 ia-3,IAM,IAM,,,,,http://scal-compass.github.io/compliance-trestle/schemas/oscal/cd
```
butler54, line 3: is this a semantically correct change? service vs Service? Onyl reason why I ask is the absolute sprawl it has throughout the tests.

tests/data/json/simple_test_profile.json
```diff
@@ -56,74 +54,87 @@
...
-                "Organization-level",
-                "System-level"
+            "Organization-level",
+            "System-level"
           ]
         },
-        {  "param_id": "ac-1_prm_3",
-            "values": [
-              "officer"
-            ]
```
butler54, line 66: There are a number of test profiles here where changes are purely formatting. Was this intended?

tests/data/json/simple_test_profile2.json
```diff
@@ -56,19 +54,21 @@
             }
           ],
           "values": [
-                "Organization-level",
-                "System-level"
+            "Organization-level",
+            "System-level"
```
butler54, line 58: Similar formatting only change here.

tests/data/csv/component-definitions/bp/component-definition.json
```diff
@@ -10,7 +10,7 @@
     "components": [
       {
         "uuid": "e2d38401-9536-4481-91d5-6331f2b88c86",
-        "type": "Service",
+        "type": "service",
```
butler54, line 13: Just adding to this a number of the other declarations are still capitalized.

vikas-agarwal76 (dismissed):
LGTM

butler54 (commented):
Overlal it looks good. - definitely braking changes - but there are a pile of 'nits' around formatting etc where (i'm guessing) AI runs

tests/data/csv/bp.sample.v2.csv
degenaro: The values are specified in lower case: [https://pages.nist.gov/OSCAL-Reference/models/v1.2.1/component-definition/json-reference/#/component-definition/components/type](https://pages.nist.gov/OSCAL-Reference/models/v1.2.1/component-definition/json-reference/#/component-definition/components/type)

degenaro: Delivered fix to require lower case for component type per OSCAL schema

tests/data/json/simple_test_profile2.json
degenaro: This is purposeful. Consistent formatting. Could be moved to a separate PR, but why? On a separate note, also added were some test cases to improve test coverage. 

Since we are going to trestle V5, I saw fit to add these clean-ups and improvements.

tests/data/csv/component-definitions/bp/component-definition.json
degenaro: Same as above: The values are specified in lower case: [https://pages.nist.gov/OSCAL-Reference/models/v1.2.1/component-definition/json-reference/#/component-definition/components/type](https://pages.nist.gov/OSCAL-Reference/models/v1.2.1/component-definition/json-reference/#/component-definition/components/type)

vikas-agarwal76 (approved):
LGTM

Issue #2037, opened 2026-01-14
trestle support for OSCAL document signing


degenaro:
## Issue description / feature objectives

The objective is to provide attestation for OSCAL documents.

#### 1. Define the signing goals and trust model

Be explicit about why and how OSCAL artifacts are signed.

What guarantees are required?

- Integrity (tamper detection)
- Authenticity (who produced it)
- Non-repudiation

Who are the trusted signers?

- CI system
- Assessment tools
- Human reviewers

Where will verification happen?

- In OSCAL Compass tools
- In downstream consumers (e.g., GRC tools, agencies)

#### 2. Choose signing and envelope standards

Choose signing methodology:

- Detached signatures (preferred?)
    -  keeps OSCAL documents unchanged
    - works across JSON/XML/YAML

Choose signing standards, some possibilities:

- Sigstore (cosign) – modern, CI-friendly, keyless support
- DSSE (Dead Simple Signing Envelope) – payload-agnostic
- JSON Web Signature (JWS) – if staying strictly JSON
- XML DSig – only if XML canonicalization is required

#### 3. Canonicalization of OSCAL documents

Define canonicalization rules per format:

- JSON: stable ordering, whitespace normalization
- YAML: convert to canonical JSON before signing
- XML: use XML Canonicalization (C14N)

Implement a normalization layer so the same OSCAL content always produces the same hash. This is critical for reproducible verification.

See issue #2013.

#### 4. Implement signing during artifact generation

This can be done as:

- enhancement to existing trestle commands
- new independent trestle signing and verification commands (preferred?)

Integrate signing into Compass workflows that produce OSCAL artifacts.

Where to sign

After:

- OSCAL generation
- Validation
- Optional policy checks

Before:

- Publishing
- Export
- Storage in artifact repositories

Implementation steps:

- Hash the canonical OSCAL document
- Create a signature envelope
- Store signature alongside the artifact
- 
```
assessment-results.json
assessment-results.sig
```
#### 5. Key management and identity

Signing is only as strong as key handling.

Possible Options:

- Key-based signing (X.509 or PGP)
- Keyless signing (OIDC identity via Sigstore)

Requirements:

- Secure storage of private keys (HSM, KMS, Vault)
- Rotation and revocation strategy
- Identity binding:
    - CI job identity
    - Tool name and version
    - Organization

#### 6. Verification support

Verification features:

- Validate signature against artifact hash
- Verify signer identity
[...]

PR #2302, opened 2026-07-27
feat: add automatic OSCAL manifest generation

MatteoFari:
## Types of changes

## Summary
Adds the beta `trestle generate-manifest` command, which creates JSON package manifests from OSCAL SSPs, profiles, component definitions, and catalogs.

The command recursively follows local OSCAL dependencies, resolves `trestle://` and back-matter references, and accepts additional documents through `--include`. Generated manifests work with `sign-manifest` and `verify-manifest`. 
## Key links:

- part of #2037

# Before you merge


MatteoFari:
@degenaro extended the command to support all eight top-level OSCAL models. HTTPS and SFTP references are downloaded through Trestle’s existing `FetcherFactory`, validated as OSCAL JSON, canonicalized, and stored under the package’s `remote` directory. Also added `get_cached_path()` to expose fetched files.

degenaro:
Here is Claude's review:

image

degenaro:
might be nice to have trestle beta enable/disable all.

degenaro:
trestle.core.commands.generate_manifest:95 ERROR: Error while generating package manifest: Signing manifest output path already exists: /home/degenaro/git/compliance-offerings.develop/trestle.ws/catalogs/ADGM-DPR-2021-Guidance/signing-mnaifest.json

might be nice to have --replace or --overwrite.

degenaro:
why does verify use --public-key PUBLIC_KEY but sign use --key KEY? Why not --private-key PRIVATE_KEY?

degenaro:
text does not line up:

image

The word `Generate `and the word `Sign` are misaligned.

degenaro:
I created this repo: https://github.com/oscal-compass-sandbox/test-signing

When I run `trestle verify -tr trestle.ws --public-key public.pem -f trestle.ws/catalogs/nist-800-53-rev4-low/manifest.json --signature trestle.ws/catalogs/nist-800-53-rev4-low/signature.json` no error is returned.

Then I modify trestle.ws/catalogs/nist-800-53-rev4-low/catalog.json by changing the version from 1.1.1 to 1.1.2.

Subsequently, the above command does not return any error? Question: a change to the catalog has no effect on verify of the corresponding manifest?

MatteoFari:
> might be nice to have trestle beta enable/disable all.

Would you prefer this in a separate PR to keep this one focused?

The --key name was an initial general design choice. I agree that `--private-key` is clearer, since that affects both sign and sign-manifest, I can address it in a separate PR as well.

> Question: a change to the catalog has no effect on verify of the corresponding manifest?

`trestle verify` verifies manifest.json as a single JSON artifact, so it does not inspect the files listed inside it. Package signing and verification require `trestle sign-manifest `.

With `verify-manifest`, modifying catalog.json should cause verification to fail because its digest no longer matches the signed package Statement.

MatteoFari:
Added resource limits for recursive HTTPS/SFTP dependency fetching and support overwriting with a flag. Signing and verification options are now standardized as --private-key and --public-key.

degenaro:
According to Bob: 

image

MatteoFari:
@degenaro added `--overwrite` to both sign and sign-manifest and updated both PR branches

degenaro:
Hi @MatteoFari. I asked Bob to review this time. He think he found a few issues. Please have a look. Only fix those that make sense to fix relative to your code. Thx!

## Bob's Findings

### 🔴 Major

#### 1. SFTP temporary file leaks on error — no cleanup in `finally`

**Location:** `trestle/core/remote/cache.py` lines 522–537 (`SFTPFetcher._do_fetch`)

The `NamedTemporaryFile` is created at line 522 with `delete=False`. The inner `try/finally` at lines 529–537 only guards the `sftp_client.get()` call. If `sftp_client.stat()` (line 519) or `_enforce_max_size()` (line 521) raises before the inner `try` is entered, `temporary_path` is never cleaned up. The outer `except Exception` at line 538 re-raises as `TrestleError` but never deletes the temp file.

Compare with the HTTPS path: `_write_limited_response()` at lines 388–402 correctly wraps the entire body in a single `try/finally` that always unlinks `temporary_path`. The SFTP branch needs the same top-level guard.

```python
# current — temporary_path can leak if stat() or enforce raises before inner try
with tempfile.NamedTemporaryFile(dir=localpath.parent, delete=False) as temporary_file:  # line 522
    temporary_path = pathlib.Path(temporary_file.name)                                   # line 523

def _enforce_transfer_size(...): ...                                                       # line 525

try:                                                                                      # line 529
    sftp_client.get(...)
    temporary_path.replace(localpath)                                                     # line 534
finally:
    if temporary_path.exists():
        temporary_path.unlink()                                                           # line 537

# fix — wrap everything from NamedTemporaryFile creation in a top-level try/finally
temporary_path: Optional[pathlib.Path] = None
try:
    with tempfile.NamedTemporaryFile(...) as temporary_file:
        temporary_path = pathlib.Path(temporary_file.name)
    sftp_client.get(..., callback=_enforce_transfer_size)
    temporary_path.replace(localpath)
finally:
    if temporary_path and temporary_path.exists():
        temporary_path.unlink()
```

---

#### 2. Post-replace `finally` can delete the output file on Windows

**Location:**
- `trestle/core/signing.py` lines 130–141 (`write_dsse_envelope`)
- `trestle/core/signing_manifest_discovery.py` lines 381–392 (`write_signing_manifest`)
- `trestle/core/remote/cache.py` lines 388–402 (`_write_limited_response`) — same pattern

All three sites follow this structure:

```python
temporary_path.replace(output_path)   # line 138 / 389 / 399
# ... falls into finally ...
finally:
    if temporary_path and temporary_path.exists():   # line 140 / 391 / 401
        temporary_path.unlink()                      # line 141 / 392 / 402
```

On POSIX, `Path.replace()` atomically renames the source and removes it, so `temporary_path.exists()` returns `False` and `unlink()` is not called. On Windows, the rename is not guaranteed to be atomic and the OS may leave the source path intact, so `temporary_path.exists()` returns `True` and the `finally` block deletes what is now the output file.

Fix: use a success flag so the `finally` only cleans up on failure.

```python
temporary_path: Optional[pathlib.Path] = None
replaced = False
try:
    with tempfile.NamedTemporaryFile(...) as f:
        temporary_path = pathlib.Path(f.name)
        f.write(serialized_envelope)
    temporary_path.replace(output_path)
    replaced = True
finally:
    if not replaced and temporary_path and temporary_path.exists():
        temporary_path.unlink()
```

---

### 🟡 Moderate

#### 3. `_enforce_transfer_size` callback double-counts the total

**Location:** `trestle/core/remote/cache.py` lines 525–527 (`SFTPFetcher._do_fetch`)

```python
def _enforce_transfer_size(transferred: int, total: int) -> None:
    self._enforce_max_size(transferred)   # line 526 — bytes transferred so far
    self._enforce_max_size(total)         # line 527 — full remote file size (redundant)
```

The Paramiko progress callback delivers `(bytes_transferred_so_far, total_file_size)`. Calling `_enforce_max_size(total)` on every tick re-checks the full declared file size, which was already checked via `sftp_client.stat()` at line 521. If the file is exactly `max_size_bytes` large, line 527 raises on the very first callback tick even though the in-progress byte count is still below the limit, making the limit exclusive instead of inclusive. The intent (guard against a growing transfer) is already covered by the post-transfer check at line 533. Remove the `_enforce_max_size(total)` call.

---

#### 4. `_run()` exception handler is unreachable / `# pragma: no cover`

**Location:** `trestle/core/commands/generate_manifest.py` lines 64–69 (`GenerateManifestCmd._run`)

```python
try:
    self.generate_manifest(...)          # line 66 — raises TrestleError on any failure
    return CmdReturnCodes.SUCCESS.value
except Exception as e:  # pragma: no cover   # line 68
    return handle_generic_command_exception(e, logger, 'Error while generating package manifest')
```

`generate_manifest()` calls `discover_signing_manifest()` and `write_signing_manifest()`, both of which raise `TrestleError` on all failure paths. Those exceptions would be caught here and turned into a clean error message — but the handler is excluded from coverage, so it is never exercised. Consider adding at least one test that forces a failure through `_run()` (e.g. missing input file) and remove the pragma.

---

#### 5. Artifact appended to `_artifacts` before children succeed

**Location:** `trestle/core/signing_manifest_discovery.py` lines 87–99 (`SigningManifestDiscovery._visit`)

```python
model_type, model = self._load_oscal_model(path)      # line 87
...
self._artifacts.append(self._manifest_artifact(path)) # line 90 ← appended here
self._visiting.append(path)                            # line 91
try:
    for href, ... in self._dependency_references(...): # line 93 — may raise
        ...
finally:
    self._visiting.pop()                               # line 97
self._visited.add(path)                                # line 98
```

If any child raises (cycle, missing file, depth limit), the parent is already counted in `_artifacts`. Because `discover_signing_manifest()` is all-or-nothing this does not produce a corrupt manifest. However, the `MAX_DISCOVERED_ARTIFACTS` guard at line 84 sees the inflated count during traversal, which can cause premature rejection of valid graphs.

---

### 🔵 Minor

#### 6. `LocalFetcher._do_fetch()` does not enforce `max_size_bytes`

**Location:** `trestle/core/remote/cache.py` lines 244–246 (`LocalFetcher._do_fetch`), line 629 (`FetcherFactory.get_fetcher`)

`_do_fetch()` is a no-op (`pass`). `max_size_bytes` is stored in the base class and enforced via `get_cached_path()`, but `get_raw()` and `get_oscal()` bypass `get_cached_path()` and read the file directly without size enforcement. Add a note in the docstring or enforce the limit consistently.

---

#### 7. HTTPS streaming only enabled when `max_size_bytes` is set

**Location:** `trestle/core/remote/cache.py` line 361 (`HTTPSFetcher._do_fetch`)

```python
stream=self._max_size_bytes is not None,   # line 361
```

When `max_size_bytes` is `None` (legacy callers), the full response body is buffered into memory via `response.text` — potentially unbounded. All discovery-initiated fetches always pass `max_size_bytes`, so the new feature is unaffected, but pre-existing call paths remain unprotected. Consider defaulting `stream=True` unconditionally.

---

### 📝 Notes

#### 8. Remote artifact filenames are opaque SHA-256 digests

**Location:** `trestle/core/signing_manifest_discovery.py` lines 215–227 (`_remote_artifact_path`)

Vendored remote artifacts are stored as `remote/<sha256(uri)>.json`, making them impossible to identify without cross-referencing the manifest. A path derived from the URI hostname and path (e.g. `remote/example.com/catalogs/catalog.json`) would make the package directory auditable without tooling. UX concern only.

---

#### 9. `ManifestArtifact` constructed with `name` for both first and second positional args

**Location:** `trestle/core/signing_manifest_discovery.py` lines 285–287 (`_manifest_artifact`)

```python
def _manifest_artifact(self, path: pathlib.Path) -> ManifestArtifact:
    name = self._artifact_name(path)
    return ManifestArtifact(name, name, OSCAL_JSON_MEDIA_TYPE, path)  # name passed twice
```

Worth verifying that the `ManifestArtifact` dataclass intends those two fields to be identical, or whether this is a copy/paste oversight.

---

#### 10. `--include` comma-separated form is non-idiomatic for CLI paths

**Location:** `trestle/core/commands/generate_manifest.py` lines 51–53, 81

`--include` accepts a comma-separated string split via `comma_sep_to_list()`. Paths containing commas would silently break. Using `action='append'` and allowing `--include` to be repeated is the standard argparse pattern for multi-value path arguments.

---

## Overall Assessment

Well-architected, security-conscious addition with strong test coverage. The two **major** issues — SFTP temp-file leak and the Windows post-replace delete race (present in all three write helpers) — should be fixed before merging. The three **moderate** issues are worth addressing in this PR or as immediate follow-ups. The remaining items are low-risk suggestions.

**Recommendation: Request changes**

MatteoFari:
@degenaro I reviewed the concerns and no SFTP leak occurs in the described case: stat() and size validation run before the temporary file exists, while all subsequent operations are covered by finally.

Regarding the windows issue, Path.replace() renames the temporary source to the destination, cleanup only targets the original temporary path, so it cannot delete the output.

3–5 do not expose bugs: size checks are non-accumulative, the exception handler is reachable and follows repository style, and discovery is all-or-nothing.

6–7 are broader fetcher hardening suggestions, manifest discovery already uses bounded streaming. These could be addressed in a separate PR.

8–10 are intentional design or optional UX choices.

So I don't think any change is required.

degenaro:
> @degenaro I reviewed the concerns and no SFTP leak occurs in the described case: stat() and size validation run before the temporary file exists, while all subsequent operations are covered by finally.
> 
> Regarding the windows issue, Path.replace() renames the temporary source to the destination, cleanup only targets the original temporary path, so it cannot delete the output.
> 
> 3–5 do not expose bugs: size checks are non-accumulative, the exception handler is reachable and follows repository style, and discovery is all-or-nothing.
> 
> 6–7 are broader fetcher hardening suggestions, manifest discovery already uses bounded streaming. These could be addressed in a separate PR.
> 
> 8–10 are intentional design or optional UX choices.
> 
> So I don't think any change is required.

@MatteoFari Thanks for looking!

degenaro (approved):
LGTM

tests/trestle/core/commands/sign_verify_test.py
```diff
@@ -338,7 +404,17 @@ def test_sign_rejects_output_that_matches_private_key(tmp_path: pathlib.Path, mo
...
-        ['trestle', 'sign', '-f', str(input_path), '--key', str(private_key_path), '-o', str(private_key_path)],
+        [
+            'trestle',
+            'sign',
+            '-f',
+            str(input_path),
+            '--private-key',
+            str(private_key_path),
+            '-o',
+            str(private_key_path),
```
degenaro, line 415: Is this right?? -o private_key_path

tests/trestle/core/commands/sign_verify_test.py
```diff
@@ -322,7 +376,19 @@ def test_sign_rejects_output_that_matches_input(tmp_path: pathlib.Path, monkeypa
...
+        'argv',
+        [
+            'trestle',
+            'sign',
+            '-f',
+            str(input_path),
+            '--private-key',
+            str(private_key_path),
+            '-o',
+            str(input_path),
```
degenaro, line 389: Is this right?? -o input_path

degenaro (changes requested):
Please see comments on sign_verify_test.py.

degenaro (approved):
LGTM

PR #2340, opened 2026-08-26
fix: raise minimum pydantic version to 2.11.0

### LINKED ISSUE #2339: model_validate override forwards kwargs unsupported on pydantic < 2.11
### Summary

`trestle/core/trestle_base_model.py:44-52` overrides `model_validate` and
always forwards the keyword arguments to `super().model_validate(...)`,
including `extra`, `by_alias` and `by_name` — even when the caller did not pass
them (they default to `None`).

On pydantic < 2.11.0, `BaseModel.model_validate` only accepts
`strict`, `from_attributes` and `context`; the `by_alias`, `by_name` and
`extra` keyword arguments were added in pydantic 2.11.0. On pydantic 2.0–2.10
any call to `trestle_base_model.TrestleBaseModel.model_validate` fails with

```
TypeError: BaseModel.model_validate() got an unexpected keyword argument 'extra'
```

(verified on pydantic 2.10.6).

The project declares `pydantic[email]>=2.0.0`, so the whole supported range
below 2.11 is affected: **all** `model_validate` calls fail there, not only
calls that pass the new arguments explicitly.

### Impact

- `TypeError` on every `model_validate` call in environments with pydantic
  2.0–2.10 — including calls that pass no keyword arguments at all.
- The error is not a `ValidationError`, so it bypasses the error handling in
  the `except ValidationError` branch of this same method.

### Suggested fix

Forward each keyword argument only when the caller actually passed it:

```python
try:
    kwargs: dict[str, Any] = {}
    if strict is not None:
        kwargs["strict"] = strict
    if from_attributes is not None:
        kwargs["from_attributes"] = from_attributes
    if context is not None:
        kwargs["context"] = context
    if by_alias is not None:
        kwargs["by_alias"] = by_alias
    if by_name is not None:
        kwargs["by_name"] = by_name
    return super().model_validate(obj, **kwargs)
```

This keeps the current behavior on pydantic >= 2.11 and restores the default
path on pydantic 2.0–2.10.

xyf5432:
Fix #2339

Raise the minimum supported pydantic version from `>=2.0.0` to `>=2.11.0`
in `pyproject.toml`. `by_alias`, `by_name` and `extra` were added to
`BaseModel.model_validate` in pydantic 2.11.0 (released 2025-03-27), and
`TrestleBaseModel.model_validate` forwards them to `super()`. With the lower
bound at 2.11.0 the forwarded keyword arguments are always supported.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
     "paramiko==5.0.0",
     "ruamel.yaml",
     "furl",
-    "pydantic[email]>=2.0.0",
+    "pydantic[email]>=2.11.0",
     "python-dotenv>=0.10.4",
     "python-frontmatter",
     "pywin32 >= 1.0;platform_system=='Windows'",


degenaro:
@xyf5432 Why not simply fix project.toml for pydantic >=2.11.0?

degenaro:
@xyf5432 Do you have an ID for our slack channel? [#oscal-compass-trestle-agileauthoring-c2p](https://cloud-native.slack.com/archives/C06F3PEPNBW)

degenaro:
Closes #2339.

xyf5432:
Thanks for the invite — happy to join! I tried signing up on the CNCF workspace, but self-signup there is restricted to @linuxfoundation.org addresses, so my personal email doesn't get through. If you could send me an invite directly, that would be great.

xyf5432:
By the way, you can send the invite to xyf8622@gmail.com.

degenaro:
Have you gone here https://cncf-slack.netlify.app/ to Join the CNCF Slack?

xyf5432:
Thanks for the link — I've joined the CNCF Slack and hopped into #oscal-compass-trestle-agileauthoring-c2p. My handle there is xyf8622.

degenaro (approved):
LGTM

PR #2360, opened 2026-09-11
feat: add digest algorithm agility to signing

### LINKED ISSUE #2358: [crypto] Implement algorithm agility in signing subsystem before GA graduation
**Summary**
The json-signing and json-manifest-signing beta features currently hardcode SHA-256 as the sole digest algorithm. Before either feature can be promoted to GA, the signing subsystem must be redesigned to support cryptographic algorithm agility — allowing operators to select from multiple approved algorithms without code changes.

**Background**
The signing subsystem was introduced as a beta feature and is currently gated behind trestle beta enable. Because it is not part of the stable interface, it is presently assessed as N/A for the crypto_algorithm_agility security requirement. However, that exemption expires at GA graduation — the requirement will apply in full at that point.

**Problem**
The following are hardcoded to SHA-256 with no switchable layer:

trestle/core/signing.py:39 — DIGEST_ALGORITHM = 'sha256' module-level constant
trestle/core/canonicalization.py:59 — sha256_digest_hex() function, SHA-256 only
trestle/core/signing.py:229 — explicit rejection of any non-SHA-256 predicate with a hard error
trestle/core/signing_manifest_discovery.py:210 — direct hashlib.sha256() call
The project's OSCAL data model already defines a comprehensive Algorithm enum in trestle/oscal/common.py:44 covering SHA-224, SHA-256, SHA-384, SHA-512, SHA3-224, SHA3-256, SHA3-384, and SHA3-512 — but the signing subsystem never uses it.

**Acceptance Criteria**
 DIGEST_ALGORITHM is replaced with a configurable parameter (trestle config file and/or CLI flag)
 sha256_digest_hex() is refactored into a generic digest_hex(data, algorithm) function driven by that parameter
 The hard rejection of non-SHA-256 predicates in signing.py is replaced with validation against the configured algorithm
 Supported algorithms include at minimum: SHA-256, SHA-384, SHA-512, and one SHA-3 variant (e.g. SHA3-256)
 The signing subsystem uses the existing Algorithm enum from trestle/oscal/common.py as the source of truth for valid algorithm names
 Algorithm selection is documented in the signing feature docs
 Tests exercise at least two distinct algorithms to prove agility

**References**
crypto_algorithm_agility security requirement (SHOULD)
trestle/core/beta_features.py:51-68 — beta feature registration
trestle/oscal/common.py:44-56 — existing Algorithm enum

MatteoFari:
## Types of changes

## Summary
Can now choose the hash algorithm when signing files or packages. SHA-256 is the default. Verification checks the signature, then uses the recorded algorithm unless the user requires a specific one. 
## Key links:

- Closes #2358

# Before you merge

### DIFF
--- a/docs/predicates/oscal-package/v1.md
+++ b/docs/predicates/oscal-package/v1.md
@@ -22,7 +22,7 @@ The package manifest format is defined by the [OSCAL Package Manifest v1 JSON Sc
 `trestle sign-manifest` creates a detached DSSE envelope. The envelope payload is an in-toto Statement that records:
 
 - the artifact names listed in the package manifest
-- the SHA-256 digest of each artifact's RFC 8785 canonical JSON bytes
+- the selected digest of each artifact's RFC 8785 canonical JSON bytes (SHA-256 by default)
 - the package predicate type and predicate fields described below
 
 The DSSE signature is made over the DSSE pre-authentication encoding of the Statement payload. The package manifest and listed JSON artifacts are not modified.
@@ -31,7 +31,7 @@ The raw manifest bytes are not signed directly. Trestle validates the manifest, 
 
 ## Statement
 
-The DSSE payload is a base64-encoded in-toto Statement with one subject for each package artifact:
+The DSSE payload is a base64-encoded in-toto Statement with one subject for each package artifact. This example uses the default SHA-256 algorithm:
 
 ```json
 {
@@ -72,6 +72,10 @@ The DSSE payload is a base64-encoded in-toto Statement with one subject for each
 ```
 
 ## Predicate fields
+
+The selected digest algorithm is recorded in each subject's `digest` key, not in the package predicate or input manifest. Signing uses the same algorithm for every artifact. After verifying the DSSE signature, verification automatically detects the algorithm, requiring exactly one digest per subject and the same supported algorithm across all subjects. An explicit `--digest-algorithm` instead requires that algorithm for every subject, including when a subject contains multiple digests, without falling back to another digest. Existing SHA-256 package envelopes retain their format.
+
+Supported digest keys are `sha224`, `sha256`, `sha384`, `sha512`, `sha3_224`, `sha3_256`, `sha3_384`, and `sha3_512`. See [Digest algorithms](../../tutorials/cli.md#digest-algorithms) for CLI selection.
 
 | Field             | Description                                                                                                   |
 | ----------------- | ------------------------------------------------------------------------------------------------------------- |
--- a/docs/predicates/oscal-signing/v1.md
+++ b/docs/predicates/oscal-signing/v1.md
@@ -22,7 +22,7 @@ For command usage, see the [`trestle sign` and `trestle verify` CLI documentatio
 `trestle sign` creates a detached DSSE envelope. The envelope payload is an in-toto Statement that records:
 
 - the subject file name
-- the SHA-256 digest of the RFC 8785 canonical JSON bytes
+- the selected digest of the RFC 8785 canonical JSON bytes (SHA-256 by default)
 - the predicate type and predicate fields described below
 
 The DSSE signature is made over the DSSE pre-authentication encoding of the
@@ -30,7 +30,7 @@ Statement payload. The original JSON file is not modified.
 
 ## Statement
 
-The DSSE payload is a base64-encoded in-toto Statement with exactly one subject:
+The DSSE payload is a base64-encoded in-toto Statement with exactly one subject. This example uses the default SHA-256 algorithm:
 
 ```json
 {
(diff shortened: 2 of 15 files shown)


degenaro:
Claude's review: 

What it does

Adds algorithm-agility to trestle sign/verify/sign-manifest/verify-manifest. SHA-256 stays the default; a new --digest-algorithm flag (backed by OSCAL's Algorithm enum: SHA-224/256/384/512, SHA3-224/256/384/512) lets you pick another hash for signing. Verification auto-detects the algorithm from the authenticated Statement rather than requiring you to specify it, and --digest-algorithm on verify becomes an optional restriction rather than a required input.

Core design — solid
canonicalization.py grows digest_algorithm_name, digest_hex, and parse_digest_algorithm as the single source of truth mapping the OSCAL enum ↔ in-toto digest names (SHA3-256 → sha3_256) ↔ hashlib. sha256_digest_hex is kept as a thin compatibility wrapper, so nothing else in the codebase silently breaks.
No downgrade-attack surface. This is the part I checked most carefully: in signing.py and signing_manifest.py, the digest actually recomputed for comparison always uses the algorithm read from the signed (DSSE-authenticated) Statement, never a value an attacker could inject unsigned. A user-supplied --digest-algorithm on verify only adds a constraint ("the signed algorithm must equal this"), it never substitutes the hash function used to validate the artifact. That's the right way to do this.
Manifest multi-subject handling correctly implements the two documented modes: auto-detect requires exactly one digest per subject and a consistent algorithm across all subjects; an explicit --digest-algorithm instead pins one algorithm and tolerates subjects with multiple recorded digests. The code matches the docs precisely here.
Backward compatibility is preserved — old SHA-256-only envelopes parse fine since sha256 is just another entry in the same enum-driven table.
Test coverage is good: parametrized round-trips across all 8 algorithms for both single-file and manifest flows, a mismatched---digest-algorithm verify failure case, and argparse-level rejection of bogus algorithm names (MD5, SHA-1, unknown) before any file I/O happens.
Minor nitpicks (non-blocking)
Algorithm is imported lazily inside several functions in canonicalization.py to dodge a circular import (trestle.oscal.common imports this module). That's a reasonable workaround, just a little repetitive — could centralize into one lazy accessor if it bothers anyone in review.
In signing.py's _validate_statement, digest_algorithm_name(digest_algorithm) != algorithm_name could simplify to digest_algorithm != signed_algorithm (both are Algorithm enum members) — purely cosmetic.
parse_digest_algorithm does a linear scan over 8 enum members per call — trivial cost, not worth optimizing.
Overall

This is a clean, self-contained feature with correct crypto reasoning (no fallback/downgrade path), good doc updates across all three predicate/tutorial docs, and thorough parametrized tests. I don't see any blocking issues — looks ready to merge pending normal CI/Sonar checks.

degenaro:
Here is Bob's review:

# PR #2360 Review — `feat: add digest algorithm agility to signing`

**Overall: ✅ Approve with minor notes**

All CI checks pass (lint, test on 6 matrix combinations, CodeQL, DCO). The one local test failure (`test_sign_and_verify_real_nist_800_53_catalog`) is a missing fixture file unrelated to this PR and also fails on the base branch.

---

## What the PR does

Adds a `--digest-algorithm` option to `trestle sign`, `trestle verify`, `trestle sign-manifest`, and `trestle verify-manifest`. SHA-256 remains the default, preserving full backward compatibility. Verification auto-detects the algorithm from the signed Statement after authenticating the DSSE signature; the option is only needed to explicitly constrain which algorithm is accepted.

---

## Strengths

1. **Correct security order**: Signature verification always happens before the payload is inspected or any artifact files are read. Tests like `test_verify_authenticates_before_detecting_digest_algorithm` and `test_manifest_authenticates_before_detecting_digest_algorithm` explicitly assert this ordering, which is the right thing to enforce.

2. **Clean `Algorithm` enum integration**: Using OSCAL's existing `Algorithm` enum as the type-level source-of-truth for supported algorithms (rather than a hand-rolled list) is a good design. The name transform in `digest_algorithm_name()` correctly maps `SHA-256 → sha256`, `SHA3-256 → sha3_256`, etc., and round-trips cleanly through `parse_digest_algorithm()`.

3. **Argparse `choices=` for early rejection**: Unsupported algorithm names fail at argument parsing (exit code 2, "invalid choice") before any file is accessed — the tests `test_signing_commands_reject_unsupported_digest_algorithms` and the manifest equivalent confirm this.

4. **Backward-compatible envelope format**: `test_explicit_sha256_preserves_default_envelope` and `test_explicit_sha256_preserves_default_manifest_envelope` verify that explicit `SHA-256` produces bit-identical output to the default, so old envelopes continue to verify without any migration.

5. **Parametrized test coverage across all 8 algorithms + `None`**: Both the `sign/verify` and `sign-manifest/verify-manifest` round-trip tests exercise every `Algorithm` value plus the default path.

6. **`TYPE_CHECKING` guard in `canonicalization.py`**: The deferred `Algorithm` import inside each function avoids a circular import, while the `TYPE_CHECKING` block gives IDEs and type checkers visibility. This is the standard pattern for this situation.

---

## Issues

### 🔴 Minor — `str(digest.get(algorithm_name))` in `signing.py:278`

```python
if hmac.compare_digest(str(digest.get(algorithm_name)), expected_digest):
```

When `algorithm_name` is not a key in `digest`, `digest.get(algorithm_name)` returns `None`, and `str(None)` is `'None'`. The comparison then safely returns `False` (because `'None'` won't equal any real hex digest), so the function correctly raises the error on line 281. However:

- The caller gets the generic "digest does not match" error rather than a more specific "subject does not contain a `<alg>` digest" message.
- The `str()` coercion hides a contract violation: by this point in `_validate_statement`, the predicate's `digestAlgorithm` has already been parsed and validated, so the subject *should* always have that key. An explicit check would be a cleaner guard.

This pattern pre-existed this PR (it was already in the code before this change), so it is not a regression introduced here. Still worth noting for a follow-up.

### 🟡 Cosmetic — `sha256_digest_hex` now routes through `digest_hex`

`sha256_digest_hex` now calls `digest_hex(data, Algorithm.SHA_256)` which calls `hashlib.new('sha256', data)` instead of `hashlib.sha256(data)` directly. The runtime cost is negligible and both paths produce identical output (verified by `test_digest_algorithm_preserves_sha256`), so this is fine for compatibility.

### 🟢 Documentation is thorough and accurate

The `cli.md` additions cover signing, verification, the `sign-manifest`/`verify-manifest` equivalents, and the `--digest-algorithm` option's semantics (select hash, not key algorithm; auto-detect on verify). The predicate spec docs are updated consistently.

---

## Summary

| Area | Finding |
|---|---|
| Security order (sign-then-verify-then-read) | ✅ Correct |
| Backward compat (SHA-256 default / no flag) | ✅ Verified by dedicated tests |
| Algorithm name mapping | ✅ Correct for all 8 OSCAL algorithms |
| CLI argument validation | ✅ Early rejection via `choices=` |
| Test coverage | ✅ Full parametrized round-trip + rejection tests |
| `str(digest.get(...))` fragility | ⚠️ Pre-existing minor issue, not a regression |
| Docs | ✅ Accurate and complete |

degenaro:
@MatteoFari Can you take a look at the minor and cosmetic issues? It's OK to say leave as is or fix in another issue (if son please open one). Once we make that determination, we can proceed to deliver this change into develop.

And thanks for doing this so quickly!!!

MatteoFari:
Yes @degenaro, I included the changes in this PR since the minor issue was consistent with its scope. Verification now explicitly checks for missing or non-string digests. I suggest leaving the cosmetic items unchanged.

degenaro (approved):
LGTM

